PR #1, opened 2022-09-01
fixed missing parameters being passed to send in Signup and Signin methods.

TimothyStiles:
parameters for signing in and signing up weren't being passed in the request to database. Fixed that and also included some unrelated comments that I was working on it the same file when I discovered this bug.

### DIFF
--- a/db.go
+++ b/db.go
@@ -4,10 +4,12 @@ import (
 	"strings"
 )
 
+// DB is a client for the SurrealDB database that holds are websocket connection.
 type DB struct {
 	ws *WS
 }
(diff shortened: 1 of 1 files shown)


TimothyStiles:
> parameters for signing in and signing up weren't being passed in the request to database. Fixed that and also included some unrelated comments that I was working on it the same file when I discovered this bug.

forgot to tag you in the above @tobiemh ^

tobiemh:
Awesome thanks @TimothyStiles , awesome work 👏 !

PR #2, opened 2022-09-01
Doc strings and example tests.

TimothyStiles:
I've added doc strings for every method in `db.go` along with several example tests. I may make another docs related pull request in the future but this is a good start for now and suits my needs.

-Tim

### DIFF
--- a/db.go
+++ b/db.go
@@ -4,10 +4,12 @@ import (
 	"strings"
 )
 
+// DB is a client for the SurrealDB database that holds are websocket connection.
 type DB struct {
 	ws *WS
 }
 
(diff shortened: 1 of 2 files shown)


TimothyStiles:
> I've added doc strings for every method in `db.go` along with several example tests. I may make another docs related pull request in the future but this is a good start for now and suits my needs.
> 
> -Tim

forgot to tag you in the above again @tobiemh!

tobiemh:
Hey @TimothyStiles, there is a current conflict preventing this from being merged. I think it's because the old commit (already merged) is also contained in this pull request...

TimothyStiles:
@tobiemh fixed and ready to be merged 👍

PR #3, opened 2022-09-02
Send and receive generic records as structs

TimothyStiles:
This PR needs to be cleaned up a smidge but it has generic struct insertion, retrieval, and unmarshaling.

PR #5, opened 2022-09-16
surreal: added examples folder

ivorytoast:
This PR is meant to add an `examples/` folder that contains a `main.go` file that can be run immediately after first cloning the repo locally. I actually followed along with: https://surrealdb.com/blog/getting-started-with-surrealdb so if someone was watching the video, they could see everything the Youtuber did in Golang. Of course, the whole video is not moved over, it is about the first 10 minutes. 

I also added a `SchemalessSelect` function which returns "rows" instead of a list of maps.

Also, will update once https://github.com/surrealdb/surrealdb.go/pull/3 gets merged to reflect the latest functionality


examples/main.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+	"github.com/surrealdb/surrealdb.go"
+)
+
+// TODO: add in ability to do INFO command on database
+// TODO: should let users specify a selector other than '*' for select statements
+// TODO: set up docker container so it also allows the client to sign in
+func main() {
+
+	println("Step 1: Connect to SurrealDB")
```
AYM1607, line 13: Use `fmt.Println` since `println` is not guaranteed to stay in the language.

db.go
```diff
@@ -75,12 +75,26 @@ func (self *DB) Query(sql string, vars map[string]any) (any, error) {
 	return self.send("query", sql, vars)
 }
 
+func (self *DB) SchemalessSelect(what string) ([]map[string]interface{}, error) {
+	output, err := self.Select(what)
+	if err != nil {
+		return make([]map[string]interface{}, 0), err
```
intabulas, line 81: you can simply `return nil, err` to avoid the allocation

db.go
```diff
@@ -75,12 +75,26 @@ func (self *DB) Query(sql string, vars map[string]any) (any, error) {
...
+	}
+
+	returnArray := make([]map[string]interface{}, 0)
```
intabulas, line 84: Since you know the # of results you should at least preallocate the slice and use the range index to assign vs appending.

examples/main.go
ivorytoast: Updated

db.go
ivorytoast: Updated

db.go
ivorytoast: So, `output, err := self.Select(what)` returns an `any`. Therefore, in this case, we do not know the size of the array at line 84. This is why I kept it as 0. If there is a way around this, please let me know!

db.go
```diff
@@ -75,12 +75,26 @@ func (self *DB) Query(sql string, vars map[string]any) (any, error) {
...
+	rows := (output).([]interface{})
```
Keitio, line 85: This will crash if output is not a `[]interface{}` (we probably don't want that)

db.go
Keitio: is you swap this line with the next, you can use `len(rows)` as the third argument to `make([]map[string]any, 0, len(rows))` and this way you get the performance of preallocation with less code change (and somewhat more safe)

db.go
```diff
@@ -75,12 +75,26 @@ func (self *DB) Query(sql string, vars map[string]any) (any, error) {
 	return self.send("query", sql, vars)
 }
 
+func (self *DB) SchemalessSelect(what string) ([]map[string]interface{}, error) {
```
Keitio, line 78: instead of the many `interface{}`, you can use `any` because this is go 1.18

Keitio, line 78: If you can, having a comment on what this method is for would be great

db.go
ivorytoast: Thank you -- updated

db.go
ivorytoast: Fixed -- made the code flow better anyways -- so thanks for pointing this out

db.go
ivorytoast: Changed

ivorytoast: Added

Keitio (approved):

timpratim:
Can you please remerge this PR?

db.go
```diff
@@ -147,6 +147,27 @@ func (self *DB) Query(sql string, vars interface{}) (interface{}, error) {
 	return self.send("query", sql, vars)
 }
 
+// SchemalessSelect returns a list of rows from the database that have been added
+// to a table without a defined schema
+func (self *DB) SchemalessSelect(what string) ([]map[string]any, error) {
```
VoyTechnology, line 152: Per Go conversions this should just be `db`. `self` and `this` is not used in Go.

tobiemh:
Hi @ivorytoast thanks so much for submitting this pull request 😃 👍 - it's greatly appreciated, even after so much time has passed! Since the date of this PR, the [Golang SDK](https://github.com/surrealdb/surrealdb.go) has seen significant changes, with support for both HTTP and WebSocket connections, full support of the SurrealDB RPC methods, a binary communication protocol using CBOR, and support for embedding SurrealDB in memory and with SurrealKV coming soon. As a result, I'm going to close this PR as it conflicts with the current codebase which has diverged significantly. Feel free to add any feedback, issues, or further pull requests to the new SDK code 🚀 🎉 !

PR #7, opened 2022-09-17
Added basic http client

MrShutCo:
Added a basic wrapper struct that allows http requests to the POST /sql endpoint, returning the response a string of the json. Also added the testify library for writing unit tests


MrShutCo:
Still a WIP, need to fix up unit tests / make more robust as well as have some kind of unmarshaling of responses

httpclient/client.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+// SurrealClient is a wrapper to more easily make HTTP calls to the SurrealDB engine
+type SurrealClient struct {
```
tobiemh, line 12: Can we call this `Client`?

tobiemh (dismissed):
Could we use `http` instead of `httpclient` for the folder name?

httpclient/client.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+type Response struct {
+	Time   string `json:"time"`
+	Status string `json:"status"`
+	Result any    `json:"result"`
+}
+
+// New creates a new instance of a SurrealClient
+func New(url, db, ns, user, password string) SurrealClient {
```
tobiemh, line 30: Could we also put `ns` before `db` here?

httpclient/client.go
```diff
@@ -0,0 +1,135 @@
...
+	return SurrealClient{
+		URL:      url,
+		DB:       db,
+		NS:       ns,
```
tobiemh, line 34: Switch these lines around.

httpclient/client.go
```diff
@@ -0,0 +1,135 @@
...
+		User:     user,
+		Password: password,
```
tobiemh, line 36: Rename this to `Pass`

MrShutCo:
> Could we use `http` instead of `httpclient` for the folder name?

@tobiemh I think this might be a bit confusing, as it conflicts with the standard golang http package, and having a `http.Client` might cause further confusion. What if I copy this code out to the root `surrealdb` package and change the struct to HttpClient or something, so that it becomes `surrealdb.HttpClient` or `surrealdb.Client`?

tobiemh:
> @tobiemh I think this might be a bit confusing, as it conflicts with the standard golang http package, and having a `http.Client` might cause further confusion. What if I copy this code out to the root `surrealdb` package and change the struct to HttpClient or something, so that it becomes `surrealdb.HttpClient` or `surrealdb.Client`?

Ahhh yes, of course - good point!

intabulas:
A general comment if I may. Two things concern me, the language of xxxAll and the use of strings. My mental model is something along the lines of (using create for an example)  Create() and CreateOne(). 

My second concern is the use of string as the value param. I really would love to see this as an interface that is marshaled inside Request so we can pass structs vs marshalling before calling.

MrShutCo:
@intabulas The first one is sorta outta my control, I was just using the functions from https://github.com/surrealdb/surrealdb.py/pull/5/files which has been merged already, as well as when going to the official docs for things like https://surrealdb.com/docs/integration/http#select-all it'll bring you to the query... Though I do admit its a bit strange. 

I agree on the second point! String was just the easiest, and now that the PRs for marshalling are in, ill revisit this and see what I can do

intabulas:
@MrShutCo gotcha, totally makes sense to follow the pattern then

MrShutCo:
@intabulas @tobiemh The PR is ready for a proper review. I've put a bunch of TODO's as well for other devs to possibly improve on this code. It isn't perfect, but hopefully its enough of a base for people to improve on :)

go.mod
```diff
@@ -2,4 +2,13 @@ module github.com/surrealdb/surrealdb.go
 
 go 1.18
 
-require github.com/gorilla/websocket v1.5.0
+require (
+	github.com/gorilla/websocket v1.5.0
+    github.com/stretchr/testify v1.8.0
```
Keitio, line 7: I'm not for the use of testify, as it is a very large dependency that most people only use for basics such as equality check and nil check, which can both easily be done with a few lines of code

I can help with the code if need be

Keitio: The rest of the code seems fine to me

MrShutCo: Is there another library that has similar things? Doing things like
`
if err != nil || data != "" {
    panic(err)
}
`
Seems clunky, and requires you to think of the inverse of what should happen, and panicking doesnt always tell you why particular things have failed or not. Mind you this is just the way I saw the other code do

Keitio: Absolutely, just create a function that does that for you !
You can look at how the require.NotNil and assert.Equal functions are implemented for inspiration
Testify is not a bad library per se, but it's very heavy
And as the creators of the language (https://go-proverbs.github.io/) say, a little copying is better than a little dependency

MrShutCo: Fair. Seems a bit tedious, but I'll see what @tobiemh thinks

client.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	// TODO: verify its a valid requesttype
+	req, err := http.NewRequest(requestType, sc.URL+endpoint, bytes.NewBuffer(bodyBytes))
+	if err != nil {
+		return Response{}, err
+	}
+	req.Header.Set("NS", sc.NS)
+	req.Header.Set("DB", sc.DB)
+	req.Header.Set("Content-Type", "application/json")
```
tobiemh, line 132: Hey @MrShutCo this is changing in `1.0.0-beta.8` to `Accept` (which it should always have been).

Can we add an `Accept: application/json` header (in addition), so it works on both `1.0.0-beta.7` and `1.0.0-beta.8`?

go.mod
tobiemh: Would be best to keep dependencies to a minimum 👍if pos

client.go
```diff
@@ -0,0 +1,153 @@
...
+//
+// SELECT * FROM type::thing(:table, :id);
+func (sc Client) SelectOne(table string, id string) (Response, error) {
+	return sc.Request(fmt.Sprintf("/key/%s/%s", table, id), "GET", "")
+}
+
+// ReplaceOne calls the endpoint PUT /key/:table/:id, executing the statement
+//
+// UPDATE type::thing($table, $id) CONTENT $body;
+func (sc Client) ReplaceOne(table, id string, body interface{}) (Response, error) {
```
tobiemh, line 81: The thing is, this is actually a replace or upsert...

client.go
```diff
@@ -0,0 +1,153 @@
...
+	return sc.Request(fmt.Sprintf("/key/%s/%s", table, id), "PUT", body)
+}
+
+// UpsertOne calls the endpoint PUT /key/:table/:id, executing the statement
+//
+// UPDATE type::thing($table, $id) MERGE $body;
+func (sc Client) UpsertOne(table, id string, body interface{}) (Response, error) {
```
tobiemh, line 88: And this is actually a merge or upsert...

client.go
```diff
@@ -0,0 +1,153 @@
...
+//
+// CREATE type::table($table) CONTENT $body;
+func (sc Client) CreateOne(table, id, body interface{}) (Response, error) {
+	return sc.Request(fmt.Sprintf("/key/%s/%s", table, id), "POST", body)
+}
+
+// CreateAll calls the endpoint POST /key/:table, executing the statement
+//
+// CREATE type::thing($table, $id) CONTENT $body;
+func (sc Client) CreateAll(table string, body interface{}) (Response, error) {
```
tobiemh, line 60: Yeah this might make more sense as just `Create`

go.mod
MrShutCo: Sounds good. I pulled in the IsEqual and removed the bells and whistles. I still believe this is better than manual error checking, and should be done in the other tests, keeping the code more readable and short.

client.go
MrShutCo: Im not sure I understand... Is this code doing the wrong thing, is the comment above incorrect or is the function named incorrectly?

client.go
MrShutCo: Ditto the above comment

client.go
tobiemh: Only that the name `Replace` implies that there needs to be something existing before replacing it. However the function actually creates or replaces the record content.

client.go
tobiemh: And this actually creates / updates / merges the record content.

client.go
MrShutCo: I was just using the naming that was in the python lib at https://github.com/surrealdb/surrealdb.py/blob/main/surrealdb/clients/http.py. If we should change it here then we should also change it in that lib too to avoid confusion. The comment about what surrealQL query is taken from the official docs too

MrShutCo: @tobiemh any updates on this? It'd be nice to get merged so others can improve on it

tobiemh:
Hi @MrShutCo just looking at this again, and thinking about the upcoming Rust library, what is the need to have an HTTP/REST specific functionality available to the end user?

With regards to the Rust library (and the latest changes to this Golang library), the HTTP/WebSocket implementations are hidden from the user.

What are your thoughts/reasoning?

MrShutCo:
@tobiemh I was just following what was being done on other repos like the Python one. If we're not exposing the endpoints and Websockets directly then this PR has no real use, and I must have missed some of the discussion on how it would be used

tobiemh:
Hey @MrShutCo ah no don't worry you haven't missed the discussion. I'm going to keep this PR open for the moment, but if we follow the Rust library (coming soon), then the HTTP endpoints will be used behind the scenes for `import` and `export` functionality. All other querying will happen over the WebSocket connection.

phughk:
Heya! Checking out open changes - I think there is still some work to be done in this area, so can't quite approve yet, but we will get back to it. Thanks for the work and patience!

phughk:
Hi @MrShutCo ! Would you be able to merge the latest changes so we can get this in? It would be really cool to have in included!

MrShutCo:
@phughk Pulled in the latest changes. It still says requested changes from @tobiemh but I cant re-request a review. Has anything changed in the http endpoints since i first made this pr?

phughk:
Landing this, thank you for persistence @MrShutCo !

phughk (commented):
Overall approve, 1 comment for many lines, tagging @tobiemh to approve or remove request for change as I can't override.

phughk:
Waiting for asserts to replace println then will approve @MrShutCo

MrShutCo:
@phughk What should be done about the response object? Ive tried to replace the fmt statements but its kinda hard since the objects look like `interface{}([]interface{}{map[string]interface{}{"child": interface{}(nil), "id": "", "name": "foobar"}}` which looks horrifying. I recall there was some discussion about this when the PR first went up, but whats the consensus now? I cant imagine anyone wanting to use that to try and unmarshal

phughk:
We can use the complex types for now and address in the future but agree

timpratim:
Hey @MrShutCo , could you please add in a new PR to skip all the new code updates? Or else merge this PR?

tobiemh:
Hi @MrShutCo thanks so much for submitting this pull request 😃 👍 - it's greatly appreciated, even after so much time has passed! Since the date of this PR, the [Golang SDK](https://github.com/surrealdb/surrealdb.go) has seen significant changes, with support for both HTTP and WebSocket connections, full support of the SurrealDB RPC methods, a binary communication protocol using CBOR, and support for embedding SurrealDB in memory and with SurrealKV coming soon. As a result, I'm going to close this PR as it conflicts with the current codebase which has diverged significantly. Feel free for any feedback, issues, or further pull requests to the new SDK code 🚀 🎉 !

PR #13, opened 2022-09-20
Fix godocs

dbut2:
rename package and fix compile errors to resolve examples not showing up in godocs. tested locally

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -1,10 +1,14 @@
-package surrealdb
+package surrealdb_test
 
-import "fmt"
+import (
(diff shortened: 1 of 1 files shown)


tobiemh:
Hey @dbut2 what's the benefit to this?

dbut2:
@tobiemh This will allow these examples to be show in the public go docs, https://pkg.go.dev/github.com/surrealdb/surrealdb.go

Example example for `fmt.Print` https://pkg.go.dev/fmt#example-Print

PR #14, opened 2022-09-20
readme design suggestion

iKaio:
in this PR i am suggesting a simple readme design, using the core `surrealdb/surrealdb` readme as inspiration.


iKaio:
too early

Issue #20, opened 2022-09-25
Serialization/Performance


iDevelopThings:
So to start, I was digging through the code, and our issue mainly stems from the fact that we need to decode the main response, to get the RPC ID for handling the callback

I was playing with this `jsonparser` package earlier today with something else, then I realised we might be able to use it here.

I did a simple benchmark, and the performance just to get the ID is much faster
![CleanShot 2022-09-25 at 17 09 34](https://user-images.githubusercontent.com/4105581/192153577-fff4649b-b714-433b-883a-510ea1c6ea3c.png)

My idea:

We can use the jsonparser, to get the id, and then we use a struct like:
```go
type RPCRawResponse struct {
	ID       string
	Data    []byte
	Decoded *RPCResponse
}
```

This "response" can be passed down the channels and such to the end function and only decoded once.

On top of this, we can build a top level "QueryResolver" which uses a generic, and is an opt-in function/struct, which handles context, decoding, checking for errors etc
This will take the byte version and pull information out in small pieces, we can also decode once to the final desired output struct from this, rather than Unmarshal -> get id -> Marshal -> Unmarshal -> check for errors and potentially more... 

How do we feel about this?
I hacked together a crappy version of this "QueryResolver" setup in my project already, but i didn't optimise or care much for code quality as i was just playing, but it's semi elegant to use
![CleanShot 2022-09-25 at 17 17 46](https://user-images.githubusercontent.com/4105581/192153973-2324b54e-ba6d-4233-8fd6-71aa141fee6c.png)

garrison-henkle:
The only issue I have with this approach is generics. I personally like the generic approach and previously had my own fork of the driver that used generics, but I'm not sure it is really the best idea for backwards compatibility. The question of what Go version to use has was raised to Tobie a few hours ago in #17, but I would imagine he'll want the driver to be as compatible as possible to allow for easier adoption in existing projects. If he okays using newer Go versions, I think it generics should 100% be used.

[...]

PR #21, opened 2022-09-25
[WIP] Resolver and Serialisation

iDevelopThings:
The idea/solution i mentioned in #20

- Implemented RPCRawResponse/Serialisation changes
- Implemented the query resolver solution
- Started tests for resolver/serialisation

Only handled the regular Query() function so far
It's pretty hacky in some places, was a quick put together as POC

TODO: Need to clean up the panics and find better solutions to them


err.go
```diff
@@ -8,6 +8,6 @@ type PermissionError struct {
 	what string
 }
 
-func (self PermissionError) Error() string {
-	return fmt.Sprint("Unable to access record:", self.what)
+func (pe PermissionError) Error() string {
+	return fmt.Sprint("Unable to access record:", pe.what)
```
Keitio, line 12: Instead of Sprint, you can just `return "Unable to access record: " + pe.what`

query_resolver.go
```diff
@@ -0,0 +1,103 @@
...
+	response     *RPCRawResponse
+	queryResults []QueryResult[T]
+	query        string
+	params       QueryParams
+	ctx          context.Context
+}
+
+// Query creates a new query resolver
+// It would be ideal if we create a global for *DB
+func Query[T any](db *DB, ctx context.Context, query string, params ...any) *QueryResolver[T] {
```
Keitio, line 30: The context should always be the first parameter here
And creating a global db would be kind of nice to use, but horrendous to test and using multiple db connections would be near impossible

query_resolver.go
```diff
@@ -0,0 +1,103 @@
...
+	var paramsData = QueryParams{}
+	if len(params) > 0 {
+		paramsData = params[0].(QueryParams)
```
Keitio, line 33: This is extremely unsafe to do, just take a `params ...QueryParams` instead, and document that only the first one is used

query_resolver.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/goccy/go-json"
+)
+
+type QueryResult[T any] struct {
+	Result []T    `json:"result"`
+	Status string `json:"status"`
+	Time   string `json:"time"`
+}
+
+type QueryParams = map[string]any
```
Keitio, line 17: Maybe this should only be called `Map` ? It would be easier to use

query_resolver.go
```diff
@@ -0,0 +1,103 @@
...
+		query:  query,
+		params: paramsData,
+		ctx:    ctx,
+	}
+
+	return resolver.resolve(db)
+}
+
+func (r *QueryResolver[T]) resolve(db *DB) *QueryResolver[T] {
+	ctx, cancel := context.WithTimeout(r.ctx, 5*time.Second)
```
Keitio, line 46: This should not be done at the driver level, because if you want to query a lot of data on a slow connection, the driver should not get in your way
That's why we pass `ctx` as the first argument in most methods

query_resolver.go
```diff
@@ -0,0 +1,103 @@
...
+	resolver := &QueryResolver[T]{
+		query:  query,
+		params: paramsData,
+		ctx:    ctx,
+	}
+
+	return resolver.resolve(db)
+}
+
+func (r *QueryResolver[T]) resolve(db *DB) *QueryResolver[T] {
```
Keitio, line 45: The first line of this method should probably be something along the lines of 
```go
if r.err != nil {
    return r
}
```
so we don't work on an invalid object

rpc.go
```diff
@@ -12,22 +18,99 @@ func (r *RPCError) Error() string {
...
+}
+
+type RPCRawResponse struct {
+	id      string
+	Data    []byte
+	Decoded *RPCResponse
+
+	decodedError bool
+	hasError     bool
+	error        *RPCError
```
Keitio, line 48: those 3 members look strange to me, hasError should be `error != nil` IMO
and document what decodedError is

rpc.go
```diff
@@ -12,22 +18,99 @@ func (r *RPCError) Error() string {
...
+
+func (res *RPCRawResponse) Error() error {
+	if res.HasError() {
+		return res.error
+	}
+
+	return nil
+}
+
+func (res *RPCRawResponse) HasError() bool {
```
Keitio, line 59: This is wayyy too complicated for a Go error check
The way to check errors in go is `if err != nil`, so this work should be done before the user gets the result back

types.go
```diff
@@ -1,8 +1,93 @@
...
+	data.Token = result.(string)
+	data.TokenData = tokenData
+
+	return nil
+}
+
+type TokenData struct {
+	IssuedAt  int    `json:"iat"`
+	NotBefore int    `json:"nbf"`
+	ExpiresAt int    `json:"exp"`
```
Keitio, line 60: Those should be time.Time (this implies a bit more work on the driver side)

types_test.go
```diff
@@ -0,0 +1,62 @@
+package surrealdb_test
+
+import (
+	"encoding/base64"
+	"fmt"
+	"testing"
+
+	"github.com/surrealdb/surrealdb.go"
+	"github.com/test-go/testify/suite"
```
Keitio, line 9: If we could not use testify that would be great, Go has a fantastic standard library that does most of the work for us, and testify is a huge dependency (go has no concept of dev dependencies, so the entire testify and its many dependencies are pulled if you want to use this driver)

env_util.go
```diff
@@ -0,0 +1,12 @@
+package surrealdb
+
+import "os"
+
+func GetEnvOrDefault(key, defaultValue string) string {
```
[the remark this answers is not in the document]
iDevelopThings: 100% agreed, will make it private

err.go
iDevelopThings: Github Copilot 😅  will change

query_resolver.go
iDevelopThings: This is true.... maybe there should be a way to change default timeout(or disable it in place for your own?)

There's cases where RPC returns an error without an ID and go hangs because of the channels, this solves that

I feel it's super cumbersome to create new contexts when making a query 🤔

rpc.go
iDevelopThings: These are mainly just states so we know that we've processed looking for the error in the raw bytes, 
decodedError = did we try to decode it?
hasError = we decoded it & we found an error

rpc.go
iDevelopThings: Agreed, it needs some refactoring

My Idea was that we run the check once, and then we rely on the error/hasError states from the above suggestion

iDevelopThings:
Sorry for any spam 😅 

I will see if i can work through these changes tomorrow and clear some bits up + add the go build tags for 18+

query_resolver.go
Keitio: That's why you make a single context per http request/else and just use it everywhere
Not having context means clunky timeouts
If someone want to just not use context, they can make a basic wrapper to abstract it away

iDevelopThings:
@Keitio what would we do with this panic here?

We wouldn't want to cancel the context/close the ws right? 🤔 


CleanShot 2022-09-26 at 14 46 47

Keitio:
Yeah, logging the error and then a `continue` would be more appropriate

iDevelopThings:
F, I thought ready for review, would request another one, my bad...

query_resolver.go
```diff
@@ -0,0 +1,106 @@
+package surrealdb
+
+import (
+	"context"
+)
+
+type QueryResolver[T any] struct {
```
Keitio, line 7: Should this type be exported ?

query_resolver_handlers.go
```diff
@@ -0,0 +1,336 @@
...
+	"fmt"
+	"time"
+
+	"github.com/buger/jsonparser"
+	"github.com/goccy/go-json"
+)
+
+var (
+	// ErrResolvedQueryResultIsInvalid Need a better name :|
+	ErrResolvedQueryResultIsInvalid = errors.New("the result from the database response is not valid, expected an array")
```
Keitio, line 14: ErrInvalidRpcResult ? make it so that it covers more cases than just array expected, but just a general rpc result problem

Keitio (commented):
My opinion is that i find the API too large to be searchable, and that's for a simple CRUD (mostly) driver
I personally think we should strive for the simplest API that's useful because it will be better understood (thus less isssues)
I think this is mostly due to exposing internal implementation details way too much (IMO even the WS should be hidden to end users)
This also doesn't strike me as very "go-like", but that's probably a me issue

iDevelopThings:
Yeah I agree, it's quite large, but it's more of an optional/additional layer the dev can opt into to keep things a bit more simplified, for example, using .Query() and getting your individual item out of the result each time, gets boring fast, and then you write a wrapper to do it.... then every person does the same thing, it makes sense to have an option for that being handled

Things that you think shouldn't be exposed like that, can be sorted, it's just a WIP, on my personal code, i don't usually care too much about things not being exported 😅

Keitio:
Yep, no problems with that ! Was just stating my personal opinion on the current state of your branch just for you to get some (albeit very opinionated) feedback

phughk:
Hey @iDevelopThings 👋 Thanks for opening this! Do you want to continue trying to merge this or can we close this PR? Will assign @timpratim if we do decide to continue. It would be good to resolve the merge conflicts to make it easier to review.

tobiemh:
Hi @iDevelopThings thanks so much for submitting this pull request 😃 👍 - it's greatly appreciated, even after so much time has passed! Since the date of this PR, the [Golang SDK](https://github.com/surrealdb/surrealdb.go) has seen significant changes, with support for both HTTP and WebSocket connections, full support of the SurrealDB RPC methods, a binary communication protocol using CBOR, and support for embedding SurrealDB in memory and with SurrealKV coming soon. As a result, I'm going to close this PR as it conflicts with the current codebase which has diverged significantly. Feel free to add any feedback, issues, or further pull requests to the new SDK code 🚀 🎉 !

Issue #23, opened 2022-10-04
GitHub actions for linting and testing


EtienneBruines:
It'd be nice to automate the testing and linting, to make pull requests validated easier.

These things might be nice to include:
- [ ] `go fmt`
- [ ] `goimports`
- [ ] `go mod tidy`
- [ ] `go mod verify`
- [ ] `go vet`
- [ ] `staticcheck` - see https://staticcheck.io/
- [ ] `gosec` - see https://github.com/securego/gosec
- [ ] `go test`

tpotjj:
Would it be preferable to have pre-commit as well?

EtienneBruines:
Personally I'd say yes - but with Go there's (as far as I know) no predefined way of doing so, so I wouldn't really know. 

But yes, it makes sense that most things the CI/CD would test (like linting) would be part of the pre-commit.

EtienneBruines:
Good catch @ffrizzo !

PR #24, opened 2022-10-04
Feat/GitHub actions formatter

tpotjj:
There is a request for implementing GitHub Workflows #23, which is also seen in other drivers (JS & Java) and the main repository.

I wanted to start with implementing the formatter, and later the linter & tester.

There are multiple ways of doing this, depending on the future needs I might already change the approach so it would be great to get some feedback on that.

The question is, would we like to use the formatting, linter & test in multiple stages? Or would we want to run these tools before we release a new version for example?

If that's the case, I would propose a different approach.


.github/workflows/format.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+    format_golang:
+      name: Format Golang
+      runs-on: ubuntu-latest
+      steps:
+        - name: Setup Golang
+          uses: actions/setup-go@v3
+          with:
+            go-version: 1.17
```
EtienneBruines, line 14: The `go.mod` file in this repo defined `1.18`, so I guess the actions can run with that as well.

scripts/go_fmt.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+if [ "$(gofmt -s -l . | wc -l)" -gt 0 ]; then
+    gofmt -d .
```
EtienneBruines, line 4: We can do this on `./...` (instead of just `.`) to also test all subdirectories, once they exist.

EtienneBruines (changes requested):

scripts/go_fmt.sh
EtienneBruines: Can we also do that for the line above? :smile:  (in the `if` statement? )

scripts/go_fmt.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+if [ "$(gofmt -s -l . | wc -l)" -gt 0 ]; then
+    gofmt -d ./...
```
EtienneBruines, line 4: Because there's the `-s` flag (`simplify`) in the `if` statement, we should also add it here to have a consistent output.

Otherwise, we would exit with code `1` because something could be simplified, but the `gofmt -d` would not report anything. I feel like we'd need `gofmt -d -s` here.

EtienneBruines (changes requested):

scripts/go_fmt.sh
tpotjj: I've tested this in a personal project with a PR:
https://github.com/tpotjj/simplebank/actions/runs/3180595633/jobs/5184359497

The previous "." looks sufficient there, also for showing multiple formatting 'errors' in different files/subdirs.

Or, do I miss something?

EtienneBruines: :astonished: did not know that! I looked it up now: https://stackoverflow.com/a/13333931/3243814

You are absolutely correct. The difference between `gofmt` and `go fmt` :sweat_smile:

EtienneBruines: @tpotjj The `-s` would still be needed in both, though.

tpotjj: Resolved this suggestion:
Small mistake between the usage of `go fmt` && `gofmt`

scripts/go_fmt.sh
tpotjj: Resolved this suggestion:
Small mistake between the usage of `go fmt` && `gofmt`

.github/workflows/format.yml
tpotjj: Resolved suggestion:

Bumped to Go version 1.18

EtienneBruines:
Great work @tpotjj ! 

Once the `-s` is in both of the `gofmt` calls, I would be happy with this PR :smile:

EtienneBruines (approved):
LGMT :smile:

Issue #22, opened 2022-10-02
Feature Request - Compatibility with go sql


EtienneBruines:
This mostly involves these tasks:
- [x] Register the go sql driver upon import (when simply calling this, but then for SurrealDB:  `import 	_ "github.com/go-sql-driver/mysql"`)
- [ ] Create a "SQL connection string" for SurrealDB - e.g. parsing the username, password, database, namespace, etc. from a string (should be relatively easy)
    - [x] username
    - [x] password
    - [x] URL
    - [ ] database name
    - [ ] namespace
- [ ] Implement the types/methods mentioned at https://pkg.go.dev/database/sql/driver - mostly around converting to/from standard Go values and SurrealDB database values.

emehrkay:
This would make using SurrealDB pretty straight forward.

emehrkay:
I'd like to help out if possible. Anything I can pick up?

emehrkay:
I've been playing around with this and I noticed a couple of things:

1. It isn't too difficult to get `FETCH` support for returned objects/structs
``` go

type fetch[T any] []T

func (yne *fetch[T]) Scan(value any) error {
	if value == nil {
		*yne = fetch[T]{}
		return nil
	}

	bytes, err := json.Marshal(value)
	if err != nil {
		return err
	}

	items := fetch[T]{}
	err = json.Unmarshal(bytes, &items)
	if err != nil {
		return err
	}

	*yne = append(*yne, items...)

	return nil
}
```

2. the fields dont necessarily come back in order, this will be kinda difficult to wrangle when it is time to Scan the data into the values

```
SELECT name, email, age, ->buys->item as bought from user:1  fetch bought
///
[
  {
    "time": "101.532µs",
    "status": "OK",
    "result": [
      {
        "age": null,
        "bought": [
          {
            "id": "item:1",
            "name": "some item",
            "price": "999.99"
          }
        ],
        "email": null,
        "name": "SurrealDB"
      }
    ]
  }
]
```

mumoshu:
Just discovered this issue 👋 
If anyone's around, can you clarify "why" we need this, rather than "how" we could implement it?

[...]

PR #28, opened 2022-10-18
feat: database/sql initial framework

EtienneBruines:
See #22

Current state: 
- [x] driver.Driver
- [x] driver.Conn
- [x] driver.Stmt
- [x] driver.Rows

The goal is to make the go_sql example run without crashing.

(Creating the PR now to get community feedback.)


sql/scanner.go
```diff
@@ -0,0 +1,121 @@
...
+	if !ok {
+		return fmt.Errorf("unsupported value")
+	}
+
+	if len(arr) == 0 {
+		return nil
+	}
+
+	for _, elem := range arr {
+		switch num := elem.(type) {
```
ffrizzo, line 46: Which is the minimum Golang version library should support? Asking because depending on this we can use generics from Golang.

EtienneBruines: Good observation!

Should probably work, version-wise. Especially since there won't be any old projects wanting to utilize this library, we could probably rely on Go 1.19+

(Go 1.18 was defined previously, which should also have generics, if I remember correctly.)

MatthewAry:
I'm learning Golang and I've been playing around with this PR with https://github.com/jmoiron/sqlx 
I've noticed that things get pretty tricky when I try to work with Arrays and Object data types. I see that [github.com/lib/pq](https://github.com/lib/pq) has a few helper functions (types?) for handling arrays such as [BoolArray](https://pkg.go.dev/github.com/lib/pq#BoolArray) through [StringArray](https://pkg.go.dev/github.com/lib/pq#StringArray) to just name a few. Would it be possible to add similar types to this PR? Perhaps update the example to show how to work with non-primitive types such as objects, geometry, arrays, etc.

MatthewAry:
It seems that I am unable to pass parameters in.
Example:
```
q.Exec("CREATE user SET name = $1, email = $2", "John", "john@gmail.com")
```
Would fail with an error `"sql: expected 0 arguments, got 2"`

Could be related to https://github.com/EtienneBruines/surrealdb.go/blob/database-sql/sql/stmt.go#L24 which I believe is used to determine how many parameters are expected for the query.

MatthewAry:
Last post on this. I think this PR should be converted to a Draft PR.

sql/scanner.go
phughk: Hey @ffrizzo @EtienneBruines - good catch! I have added generics already, so we can safely use them IMO.

phughk (commented):
Hey Etienne!

Leaving some sporadic comments, but I realised while reviewing this that actually these changes are going to be quite big! Do you think you could put together a design/issue description explaining the rationale and then we can perhaps work on smaller changes to get this through? I think it will be easier for both of us that way haha.

Thanks for the hard work btw! :)

timpratim:
Closing as this is stale, but definitely needed.

PR #29, opened 2022-10-18
Add support to run tests on Github Actions

ffrizzo:
Currently, the library has a few tests but they are not validated on the CI pipeline. 

ℹ️ Also removed the `format.yml` workflow since this is redundant with the lint that uses `golangci-lint`


EtienneBruines (approved):
Once tests succeed, this LGTM.

ffrizzo:
> Once tests succeed, this LGTM and should be an easy merge.

We are good to go :)

EtienneBruines (approved):

PR #42, opened 2023-03-07
SmartQuery

ElecTwix:
SmartQuery Added with 1.18+ Generics for return desired type. 
Tests Added.
Added statusErr for checking raw Error.

### DIFF
--- a/db.go
+++ b/db.go
@@ -92,6 +92,43 @@ func UnmarshalRaw(rawData, v interface{}) (ok bool, err error) {
 	return true, nil
 }
 
+// Used for RawQuery Unmarshaling
+type RawQuery[I any] struct {
+	Status string `json:"status"`
+	Time   string `json:"time"`
+	Result I      `json:"result"`
+	Detail string `json:"detail"`
+}
+
+// SmartUnmarshal using generics for return desired type.
+// Supports both raw and normal queries.
+func SmartUnmarshal[I any](respond interface{}, wrapperError error) (data I, err error) {
+	if err := wrapperError; err != nil {
+		return data, err
+	}
+	var bytes []byte
+	if arrResp, isArr := respond.([]interface{}); len(arrResp) > 0 {
+		if dataMap, ok := arrResp[0].(map[string]interface{}); ok && isArr {
+			if _, ok := dataMap["status"]; ok {
+				if bytes, err = json.Marshal(respond); err == nil {
+					var raw []RawQuery[I]
+					if err = json.Unmarshal(bytes, &raw); err == nil {
+						if raw[0].Status != statusOK {
+							err = errors.New(fmt.Sprintf("%s: %s", raw[0].Status, raw[0].Detail))
+						}
+						data = raw[0].Result
+					}
+				}
+				return data, err
+			}
+		}
+	}
+	if bytes, err = json.Marshal(respond); err == nil {
+		err = json.Unmarshal(bytes, &data)
+	}
+	return data, err
+}
+
 // --------------------------------------------------
 // Public methods
 // --------------------------------------------------
--- a/db_test.go
+++ b/db_test.go
@@ -241,6 +241,56 @@ func TestModify(t *testing.T) {
 	assert.Equal(t, "44", (user2).(map[string]interface{})["age"])
 }
 
+func TestSmartQuery(t *testing.T) {
+	db := openConnection(t)
+	defer db.Close()
+
+	_ = signin(t, db)
+
+	_, err := db.Use("test", "test")
+	assert.NoError(t, err)
+
+	user := []testUser{{
+		Username: "electwix",
+		Password: "1234",
+	}}
+
+	// RAW query create
+	dataArr, err := surrealdb.SmartUnmarshal[[]testUser](db.Query("Create users set Username = $user, Password = $pass", map[string]interface{}{
+		"user": user[0].Username,
+		"pass": user[0].Password,
+	}))
+
+	assert.NoError(t, err)
+	assert.Equal(t, "electwix", dataArr[0].Username)
+
+	// RAW query select
+	dataArr, err = surrealdb.SmartUnmarshal[[]testUser](db.Query("Select * from $record", map[string]interface{}{
+		"record": dataArr[0].ID,
+	}))
+
+	assert.Equal(t, "electwix", dataArr[0].Username)
+	assert.NoError(t, err)
+
+	// Select
+	data, err := surrealdb.SmartUnmarshal[testUser](db.Select(dataArr[0].ID))
+
+	assert.NoError(t, err)
+	assert.Equal(t, "electwix", data.Username)
+
+	// Select array
+	dataArr, err = surrealdb.SmartUnmarshal[[]testUser](db.Select("users"))
+
+	assert.Equal(t, "electwix", dataArr[0].Username)
+	assert.NoError(t, err)
+
+	// Delete, if needed return can ben nullable with pointer
+	nulldata, err := surrealdb.SmartUnmarshal[*testUser](db.Delete(dataArr[0].ID))
+
+	assert.NoError(t, err)
+	assert.Equal(t, (*testUser)(nil), nulldata)
+}
+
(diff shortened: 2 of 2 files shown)


phughk (commented):
Thanks for the change @ElecTwix ! Added some comments and hopefully you can rebase off the changes we will land soon for testify

db_test.go
```diff
@@ -301,3 +301,57 @@ func TestModify(t *testing.T) {
...
+		Password: "1234",
+	}}
+
+	// RAW Query Create
+	dataArr, err := surrealdb.SmartUnMarshal[[]testUser](db.Query("Create users set Username = $user, Password = $pass", map[string]interface{}{
+		"user": user[0].Username,
+		"pass": user[0].Password,
+	}))
+
+	if err != nil {
```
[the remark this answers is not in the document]
ElecTwix: of course. I will fix the other issues and finally, rewrite the tests.

db.go
```diff
@@ -92,6 +95,46 @@ func UnmarshalRaw(rawData, v interface{}) (ok bool, err error) {
...
+func SmartUnMarshal[I any](data interface{}, WrapperError error) (I, error) {
+	var Data I
+	err := WrapperError
+	if err != nil {
+		return Data, err
+	}
+	resp, ok := data.([]interface{})
+	var bytes []byte
+	if ok {
+		bytes, err = json.Marshal(resp[0])
```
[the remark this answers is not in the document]
ElecTwix: It is needed for RawQuery Response.
Also, there was a bug I rewrite the full function again, which needs review.


```css
Example Request Response
[map[result:[map[Password:1234 Username:electwix id:users:rrtqxteuwacoiyejlsio]] status:OK time:930.507µs]
```

db_test.go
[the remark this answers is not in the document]
ElecTwix: Done.

db.go
```diff
@@ -92,6 +92,43 @@ func UnmarshalRaw(rawData, v interface{}) (ok bool, err error) {
...
+type RawQuery[I any] struct {
+	Status string `json:"status"`
+	Time   string `json:"time"`
+	Result I      `json:"result"`
+	Detail string `json:"detail"`
+}
+
+// SmartUnmarshal using generics for return desired type.
+// Supports both raw and normal queries.
+func SmartUnmarshal[I any](respond interface{}, wrapperError error) (data I, err error) {
```
phughk, line 105: I wasn't familiar with this pattern where a returned tuple can be used as an input tuple to a function! Thanks for demonstrating.
That said, I am not sure if we would want such a design. We can have helper functions that provide this logic, but I think the `SmartUnmarshal` should itself be `(interface{}) -> (I, error)`. Then you could have `SmartUnmarshalWithQuery` that follows this signature. What do you think?
What I am trying to achieve with this feedback is thinking how would this code live
- will this be a long term alternative to Unmarshal? Preferrable? Less preferrable (considering potential golang frameworks)?
- will people find the signature intuitive?
- is this signature more or less convenient? We can see from the tests that it is very convenient there. But is that how people will be accessing this function, or will they be writing retry logic and retriable/nonretriable error handling before such deserialisation?


Would be interested to hear your (and others) thoughts around this.

db_test.go
```diff
@@ -241,6 +241,60 @@ func TestModify(t *testing.T) {
...
+	dataArr, err = surrealdb.SmartUnmarshal[[]testUser](db.Select("users"))
+
+	assert.Equal(t, "electwix", dataArr[0].Username)
+	assert.NoError(t, err)
+
+	// Delete, if needed return can ben nullable with pointer
+	nulldata, err := surrealdb.SmartUnmarshal[*testUser](db.Delete(dataArr[0].ID))
+
+	assert.NoError(t, err)
+	assert.Equal(t, (*testUser)(nil), nulldata)
```
phughk, line 295: This can be `assert.Nil(t, nulldata)`

db_test.go
```diff
@@ -241,6 +241,60 @@ func TestModify(t *testing.T) {
 	assert.Equal(t, "44", (user2).(map[string]interface{})["age"])
 }
 
+func TestSmartQuery(t *testing.T) {
```
phughk, line 244: These tests are really good, thanks for thinking of them! They can be better separated into sub tests though, so that it is known which properties they are testing (I am finding it hard to track state and reasoning, but I think I see what you are doing).

The way you can do that is with
```
func TestSmartQuery(t *testing.T) {
   someSetup()
   t.Run("can read select queries", func(t *testing.T) {
       testCodeHere()
   })
   t.Run("can read delete queries with nil", func (t *testing.T) {
       moreTestCodeHere()
   })
}
```
This still shares setup and order of execution, but it gives a clear separation into which properties of the code are being tested. Tracking state is also easier, since the subtests should only rely on the shared state (even though the ordered state affects them).

phughk (commented):
Some points for discussion :) Thanks!

db_test.go
ElecTwix: Done.

db_test.go
ElecTwix: Done.

db.go
ElecTwix: Hey, firstly I was thinking to create another function like said can take `interface{}` 
but when I used the library the biggest struggles I had was

- there was so much error checking needed, this was making code longer for no reason.
- there were two separate unmarshal, I needed one can handle both
- the  `interface{}`  that Query returns, I was passing interface to function to function for no reason.
- Generics usage was hard with raw Queries.
- I wanted to type safety like I wanted to specify what will be returned.

That why I created [SmartQuery](https://github.com/surrealdb/surrealdb.go/pull/42#top) this on mind

- One-line expression
- Will handle errors and return if there was an error.
- Will handle both Queries
- Will support generic out-of-the-box.
- Will return Type.
- Will wrap other Query

That's why _SmartUnmarshal_ input get `respond interface{}, wrapperError error` for convenience.
also, it is much cleaner with one-line expressions.


Ex.

```go
// What needed now
func Getstaff(plyarr *[]Player) (err error) {
	data, err := db.Query("select * from users where Staff != null", nil)
	if err != nil {
		return err
	}
	ok, err := surrealdb.UnmarshalRaw(data, plyarr)
	if err != nil {
		return err
	}
	if !ok {
		err = errors.New("Not Ok")
	}
}
```


```go
// After PR
func Getstaff(plyarr *[]Player) (err error) {
	plyarr, err = surrealdb.SmartUnmarshal[*[]IPlayer](db.Query("select * from users where Staff != null", nil))
}
```

it makes a difference huge when you have a code base that struct has methods for delete, gets, update, and init.
also, it can be used with Generics functions for just use one get to function, etc.

Empera0: This can provide us efficiency.

phughk: Yeah, actually thinking about it more I am starting to agree with you. I was going to say that across drivers we need to be versatile and provide a common baseline of behaviour that is consistent. But that is a lot of overhead for what we have today - your changes are more welcome and significant than such standardisation. So happy to approve and merge. Thanks @ElecTwix !

phughk (approved):
From the discussion we established that the function is very helpful and practical and if people need to use the original interface for finer-grained control - they still have it.
Merging now, thank you!

ElecTwix:
Maybe something is wrong with the lint both giving errors.
Should I put `//nolint:all`


[63a403](https://github.com/surrealdb/surrealdb.go/pull/42/commits/63a4032f175051945e937663fb07c723c2010707)
```go
// shadow: declaration of "err" shadows declaration at line 105 (govet)
if err := wrapperError; err != nil {...}
```
-----

[834f27](https://github.com/surrealdb/surrealdb.go/pull/42/commits/834f27a9e8c2782cab9a6f27b26b149c1cae21ea)
```go
// re-assignment to `err` can be replaced with `err := wrapperError` 
if err = wrapperError; err != nil {...}
```

phughk:
Yeah I think what that means is that you can do
```
if wrapperError != nil {
   return data, wrapperError
}
```
Because the reassignment probably isn't necessary.

Issue #41, opened 2023-03-07
[Feature Request] Auto Models Migration


vzool:
I'm looking for something like this:

```golang
models := []database.Model{
    &model.Person{},
    &model.Country{},
    &model.City{},
}

for _, model := range models {

    err := db.AutoMigrate(model)
    
    if err != nil {
    
          fmt.Printf("AutoMigrate(%s): %v\n", model.TableName(), err.Error())
    }
}
```
Thanks

phughk:
Hey, thank you for opening this feature request and opening the pull request! Really nice functionality.
Personally I think the drivers should be streamlined only for database access and protocol functionality.
Since this change is a quality-of-life feature, it may be better if it belonged in a separate repository.
If we merge this functionality into this repository, it means we need to main it's functionality and development long-term and potentially provide feature-parity with other drivers (which there could be many of).
@tobiemh what do you think?

ElecTwix:
It can be really useful for ORM-like usage.
> it means we need to main it's functionality and development long-term and potentially provide feature-parity with other drivers 

I agree but still, if it will be in the main repo it will be better maintained for a long time.
like, SurrelDB has aim for become the ultimate cloud database
ORM-like AutoMigrate will be on that list to become one.

For my personal preferences, I will use *automigrate.*
like [Gorm](https://github.com/go-gorm/gorm), etc. ORM libs, mostly all of them have *AutoMigrate*
such a comparison [bun](https://github.com/uptrace/bun) doesn't have *AutoMigrate* and most people are waiting for a long time, one of them is me.

On the other hand, I understand not all people may want an ORM-like system in their database driver.
still, if will be added it will impact big and it will be easier to use schemafull tables

vzool:
@phughk 
> If we merge this functionality into this repository, it means we need to main it's functionality and development long-term and potentially provide feature-parity with other drivers (which there could be many of).

I think it depends on the language, the community, and the database engine itself, how it should behave, not every language provides the same way of doing things, and a major version can break everything. 

Of course, all things can be done in any language but how much that will cost and gain? It is always a tradeoff.
Even, not every community is active like others.

[...]

PR #43, opened 2023-03-08
Add AutoMigration

vzool:
Covers `TYPE, INDEX, UNIQUE, ASSERT, VALUE, SCHEMAFULL, SCHEMALESS`
#41


phughk:
Haven't reviewed the code, but at a glance, it looks mostly ok. More pressing discussion on the issue page. Thank you so much for writing this!

db_test.go
```diff
@@ -15,6 +16,28 @@ type testUser struct {
 	ID       string
 }
 
+type testStatefulUser struct {
+	Username  string      `type:"string" unique:"Username"`
```
ElecTwix, line 20: can be better with `reflect.TypeOf()` instead of Struct Tag

Such as.
```go
    test1 := "test string"
    test2 := 30
    test3 := 1.5

    fmt.Println(reflect.TypeOf(test1)) // string
    fmt.Println(reflect.TypeOf(test2)) // int
    fmt.Println(reflect.TypeOf(test3)) // float64
```

vzool: Yes, of course, I thought about it before adding support for `INDEX, UNIQUE, ASSERT, VALUE, SCHEMAFULL, SCHEMALESS`.
It is not about types only, they are many needed descriptive options.

ElecTwix: > It is not about types only, they are many needed descriptive options.

I agree but we can use variable type as default type then if there was a type struct tag that can override the type.
Also like you said struct tags are needed but for for simpler use cases, this will be more convenient to use for end-user.

Something like this
```go
type testUser struct {
	Username string // No tag so its type will be string 
	Something int64  `type:"int"` // this will override
	SomeBool   bool  `index:"SomeBool,Something"` // They can use index and descriptive options with struct tags.
}
```

vzool: By default, if no type is provided then it will be converted to the closest possible type.

phughk:
Added the invalid tag. This isn't because it is "invalid", but because there is a discussion about driver scope in the issue attached ( #41 )

phughk:
Closing this PR as it goes beyond the scope of the drivers. The purpose of drivers is to provide access to the database with a minimal footprint. Any additional helpers and automation can be provided as external libraries.

PR #45, opened 2023-03-09
Concurrency safe websockets

plally:
The websocket library was not thread safe, 
- the websocket panics if it is written to concurrently
- because all the send functions were pulling from the same channel, responses could be discarded unintentionally or sent to the wrong sender.


internal/websocket/ws.go
```diff
@@ -69,33 +74,65 @@ func (ws *WebSocket) Close() error {
...
-	tick := time.NewTicker(ws.timeout)
-	for {
-		select {
-		case <-tick.C:
-			return nil, errors.New("timeout")
-		case res := <-ws.respChan:
-			if res.ID != id {
-				continue
-			}
+	timeout := time.After(ws.timeout)
```
Keitio, line 121: Good call

internal/websocket/ws.go
```diff
@@ -130,8 +169,11 @@ func (ws *WebSocket) initialize() {
 					// TODO need to find a proper way to log this error
 					continue
 				}
-
-				ws.respChan <- res
+				responseChan, ok := ws.getResponseChannel(fmt.Sprintf("%v", res.ID))
+				if ok {
+					responseChan <- res
+					close(responseChan)
+				}
```
Keitio, line 176: I think we should have a TODO here to remind us to log in an else clause, because this would definitely not be normal

internal/websocket/ws.go
```diff
@@ -18,11 +20,14 @@ const (
...
 type WebSocket struct {
-	conn    *websocket.Conn
-	timeout time.Duration
+	conn     *websocket.Conn
+	connLock sync.Mutex
+	timeout  time.Duration
 
-	respChan chan RPCResponse
-	close    chan int
+	responseChannels     map[string]chan RPCResponse
```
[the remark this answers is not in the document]
plally: According to the docs sync.Map is optimized for `(1) when the entry for a given key is only ever written once but read many times or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys`. 

We only write once then read twice per ID so im not sure if sync.Map would end up being more performant. We also lose some type safety by using sync.Map.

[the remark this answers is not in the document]
Keitio: I think this falls under case 2, but as you said we lose type safety and I don't think the performance gains are even measurable on normal workloads so do as you want

internal/rand/rand.go
```diff
@@ -2,19 +2,23 @@ package rand
...
 	"math/rand"
+	"sync"
 	"time"
 )
 
 const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
 
 var seededRand *rand.Rand = rand.New( //nolint:gosec
 	rand.NewSource(time.Now().UnixNano()))
+var seededRandLock sync.Mutex
```
plally, line 13: I added a Mutex here because random number sources are not concurrency safe and have the potential to panic when called in multiple goroutines. We were using rand.Read before which uses a Mutex internally.

internal/websocket/ws.go
```diff
@@ -134,8 +179,13 @@ func (ws *WebSocket) initialize() {
 					// TODO need to find a proper way to log this error
 					continue
 				}
-
-				ws.respChan <- res
+				responseChan, ok := ws.getResponseChannel(fmt.Sprintf("%v", res.ID))
+				if ok {
```
ffrizzo, line 183: I would suggest inversing the if with `!ok` and getting ride of else doing return first approach

phughk (approved):

PR #36, opened 2022-11-11
Update WebSocket id generation

ffrizzo:
Websocket id now is generate by the WebSocket packet removing the responsability from db

PR #47, opened 2023-03-09
Bug/id generation

plally:
When testing a with a large amount of concurrent requests for my previous PR I noticed id collisions were very frequent.
I believe this was happening because if 2 ids are generated at the same time they can get the same seed resulting in their IDs being the same. This PR changes the id generation to just set the seed once and use our own random number source instead of the library default.

We need a Mutex here because random sources are not concurrency safe, Internally `rand.Read` also uses a Mutex.

Another solution could be to just not set the seed every time we generate an ID.

### DIFF
--- a/id.go
+++ b/id.go
@@ -3,16 +3,21 @@ package surrealdb
 import (
 	"fmt"
 	"math/rand"
+	"sync"
 	"time"
 )
 
+var randSource = rand.New(rand.NewSource(time.Now().UnixNano()))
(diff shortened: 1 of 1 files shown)


ffrizzo:
I believe this is fixed here https://github.com/surrealdb/surrealdb.go/pull/36

plally:
Yeah, I think it would be. I didn't see your PR. Thanks.

PR #52, opened 2023-04-02
betav9 fix

ElecTwix:
Resp func modified.
---------
After beta v9, lib is not working.
Removed Array checking.

### DIFF
--- a/db.go
+++ b/db.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
-	"strings"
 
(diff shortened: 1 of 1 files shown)


ElecTwix:
db.Create Not Generating Unique ID.

phughk:
Hey @ElecTwix thanks for this! Some tests needed changing so I fixed that up
https://github.com/surrealdb/surrealdb.go/pull/54
Please could you grab those changes? We can continue on your branch.

PR #56, opened 2023-04-06
create WithTimeout func to set timeout

plally:
this creates a new WithTimeout function and removes the redundant surrealdb prefix from the option type.

Previously the timeout was unable to be set because websocket was an internal package. packages under the internal directory can not be imported by other modules and as a result I do not think you can create your own option function to set the timeout.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -27,7 +27,7 @@ type testUser struct {
 
 // getOptions returns a list of options to be used when creating a new websocket connection
 func getOptions() (options []surrealdb.Option) {
-	options = append(options, surrealdb.WithWriteCompression(true), surrealdb.WithTimeout(20*time.Second))
+	options = append(options, surrealdb.UseWriteCompression(true), surrealdb.WithTimeout(20*time.Second))
 	return
 }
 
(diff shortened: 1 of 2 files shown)


db_test.go
```diff
@@ -27,14 +27,14 @@ type testUser struct {
...
-func getOptions() (options []surrealdb.SurrealDBOption) {
-	// Set option to timeout after 20 seconds.
-	optionfunc := func(ws *websocket.WebSocket) error {
-		ws.Timeout = time.Duration(20) * time.Second
-		ws.Conn.EnableWriteCompression(true)
-		return nil
+func getOptions() (options []surrealdb.Option) {
+	optWriteCompression := surrealdb.Option{
+		WsOption: func(ws *websocket.WebSocket) error {
+			ws.Conn.EnableWriteCompression(true)
```
plally, line 33: If this is something we want to support setting I can also add an option for this.

phughk: From docs
> EnableWriteCompression enables and disables write compression of subsequent text and binary messages. This function is a noop if compression was not negotiated with the peer.

I say yes and let's enable it by default. Thank you for this @plally !

phughk (approved):
This is ready to merge now, but will wait with merging in case there is a further change for the option. Thank you @plally 🙏

PR #59, opened 2023-04-19
Benchmark

ElecTwix:
Benchmarks have been added.
update db_test.go
delete unused unit test.


db_test.go
```diff
@@ -425,6 +425,40 @@ func (s *SurrealDBTestSuite) TestSmartMarshalQuery() {
 	})
 }
 
+// BenchmarkCreate benchmarks the creation of a record
+func BenchmarkCreate(b *testing.B) {
```
phughk, line 412: We could probably have benchmarks in another file or package as they won't be run as often or under the same circumstances as unit and integration tests

db_test.go
```diff
@@ -425,6 +425,40 @@ func (s *SurrealDBTestSuite) TestSmartMarshalQuery() {
...
+	s := new(SurrealDBTestSuite)
+	s.SetT(&testing.T{})
+	s.SetupSuite()
+	user := &testUser{
+		Username: "tobi",
+		Password: "1234",
+	}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		s.db.Create(fmt.Sprintf("users:%d", i), user)
```
phughk: You should probably move string interpolation outside the benchmark. Have a slice of pre-made strings, perhaps.

db_test.go
```diff
@@ -425,6 +425,40 @@ func (s *SurrealDBTestSuite) TestSmartMarshalQuery() {
 	})
 }
 
+// BenchmarkCreate benchmarks the creation of a record
+func BenchmarkCreate(b *testing.B) {
+	s := new(SurrealDBTestSuite)
+	s.SetT(&testing.T{})
+	s.SetupSuite()
```
phughk, line 415: You wouldnt want to use a real database in this scenario because then you have network overhead and other issues involved in the benchmark. Perhaps instead use mocks? Gomock may be a good option for that. The create and other functions should be performance tested for parsing parameters and deserialising structs.

phughk (commented):
Nice! Leaving some comments, while I still see it is WIP :)

db_test.go
ElecTwix: done

db_test.go
ElecTwix: done

ElecTwix:
Diagram
image

db.go
```diff
@@ -21,46 +20,11 @@ var (
...
-
-// UseWriteCompression enables or disables write compression for internal websocket client
-func UseWriteCompression(useWriteCompression bool) Option {
-	return Option{
-		WsOption: func(ws *websocket.WebSocket) error {
-			ws.Conn.EnableWriteCompression(useWriteCompression)
-			return nil
-		},
-	}
+	ws iwebsocket.IWebSocket
```
phughk, line 23: In golang there is no convention to prefix interfaces with `I`. This goes for packages and the interface itself. Instead I would recommend package `websocket` and interface `WebSocket`. Then implementations can be different to `WebSocket`.

db_test.go
```diff
@@ -8,13 +8,15 @@ import (
...
+	"github.com/surrealdb/surrealdb.go/pkg/iwebsocket"
 )
 
 // TestDBSuite is a test s for the DB struct
 type SurrealDBTestSuite struct {
 	suite.Suite
-	db      *surrealdb.DB
-	options []surrealdb.Option
+	db *surrealdb.DB
+	ws iwebsocket.IWebSocket
```
phughk, line 19: We may want to have this tied to a specific implementation OR tied to general properties of a websocket. I think we do want the properties, but then this may be a map 
```
wsImplementations map[string]websocket.WebSocket{
   "gorilla": gorilla.Create(),
}
```
so that in the tests we can have
```
func TestConnects(t *testing.T) {
   for wsName, ws := range suite.wsImplementations {
      suite.Run(fmt.Sprintf("_%+v", wsName), func() {
         <TEST_CODE_HERE>
      })
   }
}

File: pkg/iwebsocket/iwebsocket.go [hunk 6]
```diff
@@ -0,0 +1,7 @@
+package iwebsocket
+
+type IWebSocket interface {
```
phughk, line 3: As mentioned, both package and interface should drop the `i`

phughk (commented):
Left some feedback but its nearly ready :)

db.go
ElecTwix: done.

pkg/iwebsocket/iwebsocket.go
ElecTwix: done.

db_test.go
ElecTwix: done

db_test.go
```diff
@@ -8,13 +8,16 @@ import (
...
+	"github.com/surrealdb/surrealdb.go/pkg/websocket"
 )
 
 // TestDBSuite is a test s for the DB struct
 type SurrealDBTestSuite struct {
 	suite.Suite
-	db      *surrealdb.DB
-	options []surrealdb.Option
+	db                *surrealdb.DB
+	ws                websocket.WebSocket
```
phughk, line 19: If we are testing the map of web socket implementations, then we don't want the web socket itself I think? Also the db would be tied per implementation. So actually the suite is a map of implementation name (string) to db struct that has a websocket tied with it.

db_test.go
```diff
@@ -25,23 +28,32 @@ type testUser struct {
...
-	suite.Run(t, SurrealDBSuite)
+	SurrealDBSuite.wsImplementations = make(map[string]websocket.WebSocket)
+
+	// Without options
+	SurrealDBSuite.wsImplementations["gorilla"] = gorilla.Create()
 
 	// With options
-	SurrealDBSuite.options = getOptions()
-	suite.Run(t, SurrealDBSuite)
+	SurrealDBSuite.wsImplementations["gorilla/w/opt"] = gorilla.Create().SetTimeOut(time.Minute).SetCompression(true)
```
phughk, line 39: We may want to avoid `/` for test names as it may get mistaken for paths. I would recommend `_` instead, even if it looks ugly. It's about the tests being rerunnable conveniently.

db_test.go
ElecTwix: in the web socket implementations map, we didn't connect yet,
it just created to WebSocket and applied to options
then we connect it gives us the ws interface with that we connect to db itself and save it into struct as db.

I was thinking to remove Ws and just put an index of the map.

phughk (approved):
Thanks so much!! Landing

PR #65, opened 2023-05-01
Update README.md

Icelain:
added information about smart unmarshal


naisofly:
@phughk @timpratim Any thoughts here?

phughk (approved):
Thanks! We can merge once green

PR #66, opened 2023-05-12
Logger added.

ElecTwix:
Logger Added.
It was on TODO I wanted to complete it.


phughk:
Can we please use https://github.com/sirupsen/logrus instead? It has levels and already has all the functionality we need. Thanks for adding this though!

phughk:
I have used it in production at many banks for years. It's good for production usage 👍

phughk:
I have used it in the past, but actually you are right. We should pick one of
- https://github.com/rs/zerolog
- https://github.com/uber-go/zap
- https://github.com/apex/log

ElecTwix:
> Zerolog seems to be an improvement on zap, and apex seems to be an integration on top of logging libraries. In that case I think we should use zerolog. What do you think?

Yeah, I agree I added zerolog implementation and some unit tests.

db_test.go
```diff
@@ -36,7 +36,7 @@ func TestSurrealDBSuite(t *testing.T) {
 	SurrealDBSuite.wsImplementations["gorilla"] = gorilla.Create()
 
 	// With options
-	SurrealDBSuite.wsImplementations["gorilla_opt"] = gorilla.Create().SetTimeOut(time.Minute).SetCompression(true)
```
phughk, line 39: Preferably we wouldn't have files generated during tests. Is it possible to capture this in memory instead?

pkg/gorilla/gorilla.go
```diff
@@ -189,12 +200,12 @@ func (ws *WebSocket) initialize() {
 				var res rpc.RPCResponse
 				err := ws.read(&res)
 				if err != nil {
-					// TODO need to find a proper way to log this error
+					ws.logger.Logger.Err(err)
```
phughk, line 203: Errors should always be returned instead of logged and missed

pkg/logger/logger_test.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/surrealdb/surrealdb.go/pkg/logger"
+)
+
+func TestLog(t *testing.T) {
+	templogger, err := logger.CreateLogFile("../../surrealdb.log")
```
phughk, line 12: This would be a perfect example of using an in memory logger as you can validate what was written to it without writing to files.

phughk (commented):
Great work! Some comments about errors and statefulnes

pkg/gorilla/gorilla.go
ElecTwix: I'm not where to return it is an anonymous func also it is concurrent I can create a channel to return the errors but I'm not sure that will logical.

db_test.go
ElecTwix: done.

pkg/logger/logger_test.go
ElecTwix: done.

pkg/gorilla/gorilla.go
```diff
@@ -74,6 +76,21 @@ func (ws *WebSocket) SetTimeOut(timeout time.Duration) *WebSocket {
 	return ws
 }
 
+// If path is empty it will use os.stdout/os.stderr
+func (ws *WebSocket) Logger(path string) *WebSocket {
+	var err error
+	ws.logger, err = logger.NewLogger(path)
```
phughk, line 82: This is the issue. Instead, we should have `logger.FromPath(path)` and `logger.FromBuffer(buffer)` or perhaps `logger.FromChannel(channel)`. We don't need the channel one so lets now add it, but hopefully it shows how that fits together. It great work overall though!

phughk (commented):
Great work! Left some comments

pkg/gorilla/gorilla.go
ElecTwix: I changed to a builder pattern it suits me better I think.
Also Updated the names of funds as you said.

pkg/gorilla/gorilla.go
phughk: Ah right I see... Ok lets log for now and please create an issue for this then. I will keep this unresolved until the issue is created by either of us, but it's ok to leave as is. Thanks ❤️

phughk (approved):

pkg/gorilla/gorilla.go
phughk: https://github.com/surrealdb/surrealdb.go/issues/67

PR #69, opened 2023-05-19
Test websocket concurrent #64

ElecTwix:
Close #64

PR #70, opened 2023-05-19
Goreport added.

ElecTwix:
Goreport added for code quality display

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,9 @@ The official SurrealDB library for Golang.
 [![](https://img.shields.io/badge/status-beta-ff00bb.svg?style=flat-square)](https://github.com/surrealdb/surrealdb.go) 
 [![](https://img.shields.io/badge/docs-view-44cc11.svg?style=flat-square)](https://surrealdb.com/docs/integration/libraries/golang)
 [![Go Reference](https://pkg.go.dev/badge/github.com/surrealdb/surrealdb.go.svg)](https://pkg.go.dev/github.com/surrealdb/surrealdb.go)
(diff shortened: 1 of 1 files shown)


ElecTwix:
@timpratim Thanks for your review, I think we should review this in #69, this PR just GoReport.

> 1. I feel 50 is a small number to test concurrency with and would suggest you increase the number of coroutines. To 
ensure high levels of concurrency. 100 seems like a good number.

That suits me.
> 2. Is it possible to change the 3 tests into a single test based on different scenarios?

Far as I remember we were doing unit testing, not a stress test if that's the case I can move some tests to another file.
> 4. There is a single wait group that is shared between 3 tests, this can introduce race conditions.

Unfortunately, we are using [testify](https://github.com/stretchr/testify)'s test suite and we cannot run all tests like race conditions in the unit test it needs to be cleaned up after every unit test.

timpratim:
Hi, @ElecTwix , thank you for your contribution to this PR. Could you please consider removing the websocket test from this particular PR? It would be beneficial to keep this PR focused on Goreport for better clarity in tracking changes.

Since the review comments have already been addressed in #69, it would be great if we could track the websocket test there instead. This will help reduce redundancy.

ElecTwix:
> Hi, @ElecTwix , thank you for your contribution to this PR. Could you please consider removing the websocket test from this particular PR? It would be beneficial to keep this PR focused on Goreport for better clarity in tracking changes.
> 
> Since the review comments have already been addressed in #69, it would be great if we could track the websocket test there instead. This will help reduce redundancy.

converted to draft PR when other PR has been merged then I can rebase and open the PR.

ElecTwix:
@timpratim this PR ready for review.

PR #73, opened 2023-06-02
Add function to check if error is about duplicate fields on an unique index

amleonc:
Just as the `os` module provides users with the `os.IsNotExist` function to test if the error was generated by trying to open a non existent file, regardless of its name, this commit adds the same functionality to check if the error comes from trying to insert data to a table with an unique index.

P.S. Also, minor typo fixed in a comment.


amleonc:
Nevermind, didn't add tests and also forgot that the error is composed.

Issue #33, opened 2022-10-26
Key path not found


spit4520:
I am assigning this politely to @TimothyStiles :sparkles: (this is long, but to keep it brief; if there is a roadmap for this package or formal way to submit and RFC; I will fix this and help refactor the package. I am currently using a different for of this repo that correct some other issue until the context changes get merged in, but I'd like to help! Just need a roadmap)

I have found a very cryptic error related to the json package and how its being called in this library. If you run a query like: `create $cid set foo = $foo` with the args: `{"cid":"", "foo":"baz"}` the query will execute successfully, but it will NOT unmarshal correctly.

For context: one of my service was throwing erroneous data because of a migration failure (psql) and that data was being blindly accepted by the service running surreal. A common paradigm forming is to push most of these problems to your DB instead of doing form filtering between all micro-services. My gRPC calls between the services just stated that they accepted a string, and did not try to enforce the string length. Typically you let the query fail, and take the reason it failed and generate your error/response from that and push it back. If you get swarmed with requests then :shrug: that's what rate limiting is for. This argument makes more sense when you are putting in 10->20 fields for one entry, and doing form validation everywhere is a waste

End of context,
The query was run successfully on surreal, and it didn't throw an error of "hey, so I can't put something into nothing". So it took the result of nothing and defaults, and tried to unmarshal it netting us the unhelpful message: `Key path not found` 

The error is wrapped a few times and propegated up the error system by encoding/json --> surreal.go --> my lib. It is extremely difficult to unwrap the error with that many layers of glue :sparkles: on it. Can we please wrap the error or add in a shim for a logger to catch is like in the Stripe library to make this easier to debug? This error is extremely cryptic and took me a week (60->80 hours) to solve. I was taking apart kubernetes and the rest of my tooling to make sure this wasn't my fault

phughk:
This is tied to the error codes issue we have seen; We are rethinking how errors will work for the golang driver.
https://github.com/surrealdb/surrealdb.go/pull/74

PR #74, opened 2023-06-03
Add function to check error on unique index.

pof3:
Just as the `os` module provides users with the `os.IsNotExist` function to test if the error was generated by trying to open a non existent file, regardless of its name, this commit adds the same functionality to check if the error comes from trying to insert data to a table with an unique index. This example, which is in the test generated for the function, shows its behavior:

```go
...
err1 := errors.New("sending request failed for method 'create': There was a problem with the database: Database index `userNameIdx` already contains 'Mark', with record `user:⟨2⟩`")
err2 := errors.New("sending request failed for method 'create': There was a problem with the database: Database index \"''``\" already contains '\\', with record `user:⟨{ foo = []}⟩`")
err3 := errors.New("sending request failed for method 'create': There was a problem with the database: Database index `userNameIdx` THIS MUST NOT PASS, with record `user:⟨2⟩`")
err4 := errors.New(" Database index `uniqueBlob` already contains { date: 'today', location: 'London' }, with record `foo:2`")
err5 := errors.New(" database index `uniqueBlob` already contains { date: 'today', location: 'London' }, with record `foo:2`")
err6 := errors.New("Database record `foo:1` already exists")

fmt.Println(IsDuplicateUniqueIdx(err1))
fmt.Println(IsDuplicateUniqueIdx(err2))
fmt.Println(IsDuplicateUniqueIdx(err3))
fmt.Println(IsDuplicateUniqueIdx(err4))
fmt.Println(IsDuplicateUniqueIdx(err5))
fmt.Println(IsDuplicateUniqueIdx(err6))
...
```

Will print:

```
true
true
false
true
false
false
```

### DIFF
--- a/db.go
+++ b/db.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"reflect"
+	"regexp"
 
 	"github.com/surrealdb/surrealdb.go/pkg/websocket"
 )
@@ -318,4 +319,16 @@ func isSlice(possibleSlice interface{}) bool {
 	}
 
 	return slice
+}
+
+// IsDuplicateUniqueIdx returns if the error was caused by
+// trying to create a record with a field that is duplicated
+// in an unique index. This function will return false if the
+// error was caused by a duplicated ID.
+func IsDuplicateUniqueIdx(err error) bool {
+	if err == nil {
+		return false
+	}
+	match, _ := regexp.MatchString(`Database index .* already contains .*, with record .*`, err.Error())
+	return match
 }
--- a/db.go
+++ b/db.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-
 	"reflect"
 	"regexp"
 
@@ -321,10 +320,9 @@ func isSlice(possibleSlice interface{}) bool {
 	return slice
 }
 
-// IsDuplicateUniqueIdx returns if the error was caused by
+// IsDuplicateUniqueIdx returns true if the error was caused by
 // trying to create a record with a field that is duplicated
-// in an unique index. This function will return false if the
-// error was caused by a duplicated ID.
+// in an unique index.
 func IsDuplicateUniqueIdx(err error) bool {
 	if err == nil {
 		return false
--- a/db_test.go
+++ b/db_test.go
@@ -2,6 +2,7 @@ package surrealdb_test
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"testing"
@@ -436,4 +437,58 @@ func assertContains[K fmt.Stringer](s *SurrealDBTestSuite, input []K, matcher fu
 	}
 	s.NotEmptyf(matching, "Input %+v did not contain matching element", fmt.Sprintf("%+v", input))
 	return matching
+}
+
+func TestIsDuplicateUniqueIdx(t *testing.T) {
+	type args struct {
+		err error
+	}
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{
+			name: "No special characters",
+			args: args{
+				err: errors.New("sending request failed for method 'create': There was a problem with the database: Database index `userNameIdx` already contains 'Mark', with record `user:⟨2⟩`"),
+			},
+			want: true,
+		},
+		{
+			name: "With special characters",
+			args: args{
+				err: errors.New("sending request failed for method 'create': There was a problem with the database: Database index \"''``\" already contains '\\', with record `user:⟨{ foo = []}⟩`"),
+			},
+			want: true,
+		},
+		{
+			name: "Wrong substring in the middle",
+			args: args{
+				err: errors.New("sending request failed for method 'create': There was a problem with the database: Database index `userNameIdx` THIS MUST NOT PASS, with record `user:⟨2⟩`"),
+			},
+			want: false,
+		},
+		{
+			name: "With object field",
+			args: args{
+				err: errors.New(" Database index `uniqueBlob` already contains { date: 'today', location: 'London' }, with record `foo:2`"),
+			},
+			want: true,
+		},
+		{
+			name: "Case sensitive",
+			args: args{
+				err: errors.New(" database index `uniqueBlob` already contains { date: 'today', location: 'London' }, with record `foo:2`"),
+			},
+			want: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := surrealdb.IsDuplicateUniqueIdx(tt.args.err); got != tt.want {
+				t.Errorf("IsDuplicateUniqueIdx() = %v, want %v", got, tt.want)
+			}
+		})
+	}
 }
--- a/db_test.go
+++ b/db_test.go
@@ -2,7 +2,6 @@ package surrealdb_test
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"os"
 	"testing"
@@ -11,7 +10,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	"github.com/surrealdb/surrealdb.go"
-	gorilla "github.com/surrealdb/surrealdb.go/pkg/gorilla"
+	"github.com/surrealdb/surrealdb.go/pkg/gorilla"
 	"github.com/surrealdb/surrealdb.go/pkg/logger"
 	"github.com/surrealdb/surrealdb.go/pkg/websocket"
 )
@@ -146,7 +145,7 @@ func (s *SurrealDBTestSuite) TestCreate() {
 
 	s.Run("Single create works", func() {
 		userData, err := s.db.Create("users", testUser{
-			Username: "johnny",
+			Username: "tim",
 			Password: "123",
 		})
 		s.Require().NoError(err)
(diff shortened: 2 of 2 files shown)


db_test.go
```diff
@@ -437,3 +438,57 @@ func assertContains[K fmt.Stringer](s *SurrealDBTestSuite, input []K, matcher fu
...
+			name: "No special characters",
+			args: args{
+				err: errors.New("sending request failed for method 'create': There was a problem with the database: Database index `userNameIdx` already contains 'Mark', with record `user:⟨2⟩`"),
+			},
+			want: true,
+		},
+		{
+			name: "With special characters",
+			args: args{
+				err: errors.New("sending request failed for method 'create': There was a problem with the database: Database index \"''``\" already contains '\\', with record `user:⟨{ foo = []}⟩`"),
```
ElecTwix, line 461: maybe chop the multiple lines?

db_test.go
```diff
@@ -437,3 +438,57 @@ func assertContains[K fmt.Stringer](s *SurrealDBTestSuite, input []K, matcher fu
...
+	}
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{
+			name: "No special characters",
+			args: args{
+				err: errors.New("sending request failed for method 'create': There was a problem with the database: Database index `userNameIdx` already contains 'Mark', with record `user:⟨2⟩`"),
```
ElecTwix, line 454: same as 461

db_test.go
```diff
@@ -437,3 +438,57 @@ func assertContains[K fmt.Stringer](s *SurrealDBTestSuite, input []K, matcher fu
...
+			name: "With special characters",
+			args: args{
+				err: errors.New("sending request failed for method 'create': There was a problem with the database: Database index \"''``\" already contains '\\', with record `user:⟨{ foo = []}⟩`"),
+			},
+			want: true,
+		},
+		{
+			name: "Wrong substring in the middle",
+			args: args{
+				err: errors.New("sending request failed for method 'create': There was a problem with the database: Database index `userNameIdx` THIS MUST NOT PASS, with record `user:⟨2⟩`"),
```
ElecTwix, line 468: same as 461

db_test.go
```diff
@@ -437,3 +438,57 @@ func assertContains[K fmt.Stringer](s *SurrealDBTestSuite, input []K, matcher fu
 	s.NotEmptyf(matching, "Input %+v did not contain matching element", fmt.Sprintf("%+v", input))
 	return matching
 }
+
+func TestIsDuplicateUniqueIdx(t *testing.T) {
```
ElecTwix, line 442: It can be better if we use Testify's Suit and test with the database itself for testing unexpected behavior.

db_test.go
```diff
@@ -437,3 +438,57 @@ func assertContains[K fmt.Stringer](s *SurrealDBTestSuite, input []K, matcher fu
...
+		},
+		{
+			name: "Case sensitive",
+			args: args{
+				err: errors.New(" database index `uniqueBlob` already contains { date: 'today', location: 'London' }, with record `foo:2`"),
+			},
+			want: false,
+		},
+	}
+	for _, tt := range tests {
```
ElecTwix, line 487: tt can be test or similar to that

db_test.go
```diff
@@ -437,3 +438,57 @@ func assertContains[K fmt.Stringer](s *SurrealDBTestSuite, input []K, matcher fu
 	s.NotEmptyf(matching, "Input %+v did not contain matching element", fmt.Sprintf("%+v", input))
 	return matching
 }
+
+func TestIsDuplicateUniqueIdx(t *testing.T) {
+	type args struct {
+		err error
+	}
+	tests := []struct {
```
ElecTwix, line 446: there is no need for this we can use the testify's suite like other ones

db_test.go
```diff
@@ -437,3 +438,57 @@ func assertContains[K fmt.Stringer](s *SurrealDBTestSuite, input []K, matcher fu
 	s.NotEmptyf(matching, "Input %+v did not contain matching element", fmt.Sprintf("%+v", input))
 	return matching
 }
+
+func TestIsDuplicateUniqueIdx(t *testing.T) {
+	type args struct {
```
ElecTwix, line 443: same with other struct review

ElecTwix (changes requested):
Looks promising but it needs a rewrite for the PR's unit test.

db_test.go
amleonc: Yeah, my bad, didn't really take the time to properly write the test with Testify. I'm going to work on it and send another commit. Thanks for your advice!

amleonc:
Just re wrote the tests using testify. all of them are passing.

ElecTwix:
> Just re wrote the tests using testify. all of them are passing.

you can resolve the issues has been fixed with this commit.
Nice job. :+1:

db.go
```diff
@@ -319,3 +319,14 @@ func isSlice(possibleSlice interface{}) bool {
...
 }
+
+// IsDuplicateUniqueIdx returns true if the error was caused by
+// trying to create a record with a field that is duplicated
+// in an unique index.
+func IsDuplicateUniqueIdx(err error) bool {
+	if err == nil {
+		return false
+	}
+	match, _ := regexp.MatchString(`Database index .* already contains .*, with record .*`, err.Error())
```
timpratim, line 330: ToDo:
Create ticket on SurrealDB project to include error code in error response.

db_test.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	"github.com/surrealdb/surrealdb.go"
-	gorilla "github.com/surrealdb/surrealdb.go/pkg/gorilla"
+	"github.com/surrealdb/surrealdb.go/pkg/gorilla"
```
timpratim, line 13: ToDo:

Add missing dependencies to depgaurd.

db_test.go
```diff
@@ -145,7 +145,7 @@ func (s *SurrealDBTestSuite) TestCreate() {
 
 	s.Run("Single create works", func() {
 		userData, err := s.db.Create("users", testUser{
-			Username: "johnny",
+			Username: "tim",
```
timpratim, line 148: Is this related to a test?
If not, please revert to original text.

db_test.go
```diff
@@ -156,7 +156,7 @@ func (s *SurrealDBTestSuite) TestCreate() {
 		s.Require().NoError(err)
 		s.Len(userSlice, 1)
 
-		s.Equal("johnny", userSlice[0].Username)
+		s.Equal("tim", userSlice[0].Username)
```
timpratim, line 159: Same here.

db_test.go
```diff
@@ -165,7 +165,7 @@ func (s *SurrealDBTestSuite) TestCreate() {
 		data := make([]testUser, 0)
 		data = append(data,
 			testUser{
-				Username: "johnny",
+				Username: "lu",
```
timpratim, line 168: Same here.

db_test.go
```diff
@@ -437,3 +437,89 @@ func assertContains[K fmt.Stringer](s *SurrealDBTestSuite, input []K, matcher fu
 	s.NotEmptyf(matching, "Input %+v did not contain matching element", fmt.Sprintf("%+v", input))
 	return matching
 }
+
+func (s *SurrealDBTestSuite) TestIsDuplicateUniqueIdx() {
+	query := `DEFINE INDEX uniqueNameIdx ON TABLE users COLUMNS username UNIQUE;`
+	_, err := s.db.Query(query, nil) // because of this query, I modified the 'username' field in other tests.
```
timpratim, line 443: Tests should be independent. Please declare every table and data in this test.

db_test.go
```diff
@@ -437,3 +437,89 @@ func assertContains[K fmt.Stringer](s *SurrealDBTestSuite, input []K, matcher fu
...
+
+		err = nil
+
+		_, err = s.db.Select("users:notexists")
+		s.Equal(err, surrealdb.ErrNoRow)
+		s.False(surrealdb.IsDuplicateUniqueIdx(err))
+	})
+	err = nil
+
+	// and finally, check with a tricky ID; this is the only way the function can give a false positive
```
timpratim, line 513: Can we change this to an ignored testcase? That's because we want to verify important behaviour rather than verify implementation. Also it's also ok to have to a test in the comments rather than a working test.

timpratim (commented):
The contribution is definitely helpful and looking forward to the final PR. The Todos are not for  you to work on but for us to look into more. Happy to approve once the changes have been done. 

-- Regex is known for decreasing performance and we will go ahead and add error code to make the implementation uniform across all drivers.

pof3:
Thanks, @timpratim !
Yeah, I know regexp are not an ideal solution, in Go we'd rather use the built in functions `errors.Is`, `errors.Unwrap`, etc. Nonetheless, the current driver works on top of websockets, and no further modifications can be done without immediate effects on other drivers and stacks. As you say, returning the code error at the beginning would be a nice workaround, so we can standardize it and make the proper changes all together. This PR adds a quick way to do the job, so it might be helpful for users, and I think it wouldn't be an issue for backwards compatibility, since the underlying behaviour in future implementation will be the same as expected.

phughk:
Linking this error ticket which also covers incorrect handling of error responses
https://github.com/surrealdb/surrealdb.go/issues/33
Agreed with @timpratim we are happy to go ahead with regex parsing 👍

tobiemh:
Hi @pof3 thanks so much for submitting this pull request 😃 👍 - it's greatly appreciated, even after so much time has passed! Since the date of this PR, the [Golang SDK](https://github.com/surrealdb/surrealdb.go) has seen significant changes, with support for both HTTP and WebSocket connections, full support of the SurrealDB RPC methods, a binary communication protocol using CBOR, and support for embedding SurrealDB in memory and with SurrealKV coming soon. As a result, I'm going to close this PR as it conflicts with the current codebase which has diverged significantly. Feel free to add any feedback, issues, or further pull requests to the new SDK code 🚀 🎉 !

PR #88, opened 2023-07-25
Add support  for Live Queries

Atoo35:
First of all i apologize for the worst code ever.

I am getting this segmentation fault error
image

In the above image the `data` print statement is from within `ws.read()` and I see that I have received the live query response. I think this should have its own type. Being fairly new to go and channels, i created a go routine which keeps extracting data from the connection and i believe this is where i am going wrong
I am struggling to understand the reason, but from what I imagine it could be maybe from the `ws.read()` function since when the server sends the response back to the client, I am reading the response and trying to process it and send it to the response channel created. I think I need to create new set of channels similar to responseChannels and the locks, maybe naming it to notificationChannels but that is later.


ElecTwix:
I didn't go too far with reviews because of draft PR such as print statements...
other than it looking good :)
Thanks for your contribution. :+1:

Atoo35:
@ElecTwix thanks but its not working lol

Atoo35:
Yes I am getting segmentation fault. That's what's the issue is. I did mention a few points in the description before about why that might be but not sure if that is indeed the reason or not and how to resolve those

ElecTwix:
> Yes I am getting segmentation fault. That's what's the issue is. I did mention a few points in the description before about why that might be but not sure if that is indeed the reason or not and how to resolve those

I will look into it when I had time
also in the meantime you could try to debug

Atoo35:
I think i found the error. So in the `gorilla.go` file, we have the initialize method which also listens to the socket connection and reads incoming message. This along with the same socket listen code for notifications is problematic. once a response is read, it is "lost", as in both the instances cant read the same message, randomly either the initialize or the LiveNotification code will be invoked causing the other to fail and hence segmentation fault. 

So this listening should be done centrally and the response should be handled accordingly. I am not sure how would that be done
Attaching the modified initialize function which works fine but not everytime
```golang
func (ws *WebSocket) initialize() {
	go func() {
		for {
			select {
			case <-ws.close:
				return
			default:
				var res rpc.RPCResponse
				err := ws.read(&res)
				if err != nil {
					ws.logger.Logger.Err(err)
					ws.logger.LogChannel <- err.Error()
					continue
				}
				var responseChan chan rpc.RPCResponse
				var ok bool
				if res.ID != nil && res.ID != "" {
					responseChan, ok = ws.getResponseChannel(fmt.Sprintf("%v", res.ID))
					if !ok {
						err = errors.New("ResponseChannel is not ok")
						ws.logger.Logger.Err(err)
						ws.logger.LogChannel <- err.Error()
						continue
					}
					responseChan <- res
					close(responseChan)
				} else {
					fmt.Println("res", res.Result.(map[string]interface{})["id"])
					responseChan, ok = ws.getResponseChannel(fmt.Sprintf("%v", res.Result.(map[string]interface{})["id"]))
					if !ok {
						err = errors.New("ResponseChannel is not ok")
						ws.logger.Logger.Err(err)
						ws.logger.LogChannel <- err.Error()
						continue
					}
					responseChan <- res
					// close(responseChan)
				}

			}
		}
	}()
}

```

phughk:
Hey thanks for this!
Yes, the stack trace points at the line where nil isn't handled. Either handle or return an error there. It would be a separate PR and I would like a test. I haven't looked into it, but a different design may be viable as a solution as well: where nil isn't possible, for example.

Atoo35:
@phughk i have pushed some new code. The local test worked fine for the TestLive testcase but apparently something is missing. I am not sure what the issue is. Could you please guide me on the latest code?

Atoo35:
Can someone please take a look at this? I am out of ideas now 🤣

Atoo35:
Guys lets get this closed soon please?

ElecTwix:
> Guys let's get this closed soon please?

the late time I talked with @phughk he said the data structure can be changed for live queries when beta 10, 
maybe we should for beta 10 to see those changes and implement them afterward and wait for that to happen?

Atoo35:
ahh i seee ok ok

phughk:
Heya both 👋 

We did intend to change the message format. We probably won't. It won't be a big deal if we do, as the mechanism will be essentially the same. So it is safe to program against the current format.

The stack trace can be traced to the file `gorilla.go` line `273`. I haven't looked, but I am guessing that is something uninitialised or defaulting to nil. Have a look at the cause inside the gorilla implementation.

I recommend merging with the main branch as a significant change landed that moves files around (and there are merge conflicts now). This should be done before debugging imo.

ElecTwix:
> The stack trace can be traced to the file `gorilla.go` line `273`. I haven't looked, but I am guessing that is something uninitialized or defaulting to nil. Have a look at the cause inside the gorilla implementation.

I looked into it before it was error caused because it is not killed and the RPC we made will also cause the live query to send back thing so we waited for 1 but we got 2 and that will block other RPC responds because go wait for channel to get that data, I was looking into killing live query after it but I was couldn't manage to do it I will look into it right now maybe I could manage to find, thanks for clarification also

Atoo35:
just merged the main branch

ElecTwix:
Hey @Atoo35 I created PR for the loop fix https://github.com/Atoo35/surrealdb.go/pull/1
It is working nightly.
Also, @phughk thanks for this https://github.com/surrealdb/surrealdb/pull/2402 it helped a lot

Atoo35:
@ElecTwix thanks for the help but i swear to god I tried this. why didn't it work lol. lemme try this again xD

Atoo35:
@ElecTwix you are a saviour, i tried the kill but with an older nightly where it errored out. This is good to go now!!

internal/mock/mock.go
```diff
@@ -17,6 +20,10 @@ func (w *ws) Close() error {
 	return nil
 }
 
+func (w *ws) LiveNotifications(id string) (chan rpc.RPCResponse, error) {
+	return nil, nil
```
phughk, line 24: Instead or `nil, nil` you can return an error. `nil, errors.New("live queries are unimplemented for mocks")`

pkg/gorilla/gorilla.go
```diff
@@ -206,15 +215,28 @@ func (ws *WebSocket) initialize() {
...
-					err = errors.New("ResponseChannel is not ok")
-					ws.logger.Logger.Err(err)
-					ws.logger.LogChannel <- err.Error()
-					continue
+				var responseChan chan rpc.RPCResponse
+				var ok bool
+				if res.ID != nil && res.ID != "" {
+					responseChan, ok = ws.getResponseChannel(fmt.Sprintf("%v", res.ID))
+					if !ok {
+						err = errors.New("ResponseChannel is not ok")
```
phughk, line 223:
```suggestion
						err = fmt.Errorf("unavailable ResponseChannel %+v", res.ID)
```

pkg/gorilla/gorilla.go
```diff
@@ -206,15 +215,28 @@ func (ws *WebSocket) initialize() {
...
+						ws.logger.Logger.Err(err)
+						ws.logger.LogChannel <- err.Error()
+						continue
+					}
+					responseChan <- res
+					close(responseChan)
+				} else {
+					responseChan, ok = ws.getResponseChannel(fmt.Sprintf("%v", res.Result.(map[string]interface{})["id"]))
+					if !ok {
+						err = errors.New("ResponseChannel is not ok")
```
phughk, line 233:
```suggestion
						err = fmt.Errorf("unavailable ResponseChannel %+v", resolved_id)
```

pkg/gorilla/gorilla.go
```diff
@@ -206,15 +215,28 @@ func (ws *WebSocket) initialize() {
...
+						ws.logger.Logger.Err(err)
+						ws.logger.LogChannel <- err.Error()
+						continue
+					}
+					responseChan <- res
+					close(responseChan)
+				} else {
+					responseChan, ok = ws.getResponseChannel(fmt.Sprintf("%v", res.Result.(map[string]interface{})["id"]))
```
phughk, line 231:
```suggestion
          resolved_id, ok = res.Result.(map[string]interface{})["id"]
          if !ok {
              err := errors.New("response did not contain an 'id' field")
              log.WithField("result", res.Result).Errorf("Failed to handle response", err)
              return err
          }
					responseChan, ok = ws.getResponseChannel(fmt.Sprintf("%v", res.Result.(map[string]interface{})["id"]))
```

pkg/websocket/websocket.go
```diff
@@ -1,7 +1,10 @@
 package websocket
 
+import "github.com/surrealdb/surrealdb.go/internal/rpc"
+
 type WebSocket interface {
 	Connect(url string) (WebSocket, error)
 	Send(method string, params []interface{}) (interface{}, error)
 	Close() error
+	LiveNotifications(id string) (chan rpc.RPCResponse, error)
```
phughk, line 9: The return type shouldnt be an rpc response (an implementation detail). It should be a `Notification` type, that you would need to add to the golang driver.
This is what it looks like:
https://github.com/surrealdb/surrealdb/blob/main/lib/src/dbs/notification.rs#L24-L28

db_test.go
```diff
@@ -124,6 +124,27 @@ func signin(s *SurrealDBTestSuite) interface{} {
...
+			"username": "johnny",
+			"password": "123",
+		})
+		s.Require().NoError(e)
+		notification := <-notifications
+		res := notification.Result.(map[string]interface{})
+		s.Require().Equal("CREATE", res["action"])
+		s.Require().Equal(live.(string), res["id"])
+		// Need to kill before get out from scope
+		_, err = s.db.Kill(live.(string))
```
phughk, line 144: This is awesome, nice catch!
You can call it immediately after live using `defer`.
Adding suggestion above as well. The assert can be removed from here.
```suggestion

```

db_test.go
```diff
@@ -124,6 +124,27 @@ func signin(s *SurrealDBTestSuite) interface{} {
 	return signin
 }
 
+func (s *SurrealDBTestSuite) TestLive() {
+	s.Run("Live notifications", func() {
+		live, err := s.db.Live("users")
+		s.Require().NoError(err)
```
phughk, line 130:
```suggestion
		s.Require().NoError(err)
		defer func() {
		   _, err := s.db.Kill(live.(string))
		    s.Require().NoError(err)
		}
```

phughk (commented):
Great work! Some suggested fixes :)

pkg/gorilla/gorilla.go
Atoo35: I am a bit confused about `log.WithField()`. we are not using the `log` package here, did you mean something like `ws.logger.Logger`?

ElecTwix: Yeah, I think he meant `ws.logger.Logger`
also, snake case are not usually used in go for various mostly files and constants that why golang-linter giving error.

db.go
```diff
@@ -109,6 +110,10 @@ func (db *DB) Insert(what string, data interface{}) (interface{}, error) {
 	return db.send("insert", what, data)
 }
 
+func (db *DB) LiveNotifications(id string) (chan rpc.RPCResponse, error) {
```
ElecTwix, line 113: people cannot use internal packages maybe we can move rpc to pkg

db_test.go
```diff
@@ -124,6 +124,27 @@ func signin(s *SurrealDBTestSuite) interface{} {
...
+		notifications, er := s.db.LiveNotifications(live.(string))
+		// create a user
+		s.Require().NoError(er)
+		_, e := s.db.Create("users", map[string]interface{}{
+			"username": "johnny",
+			"password": "123",
+		})
+		s.Require().NoError(e)
+		notification := <-notifications
+		res := notification.Result.(map[string]interface{})
```
ElecTwix, line 140: maybe we can do for safety and check `ok`
```suggestion
		res, ok := notification.Result.(map[string]interface{})
```

pkg/gorilla/gorilla.go
```diff
@@ -103,6 +103,16 @@ func (ws *WebSocket) Close() error {
...
 }
 
+func (ws *WebSocket) LiveNotifications(id string) (chan rpc.RPCResponse, error) {
+	c, err := ws.createResponseChannel(id)
+	if err != nil {
+		ws.logger.Logger.Err(err)
+		ws.logger.LogChannel <- err.Error()
+		return nil, err
+	}
+	return c, nil
```
ElecTwix, line 113: we can make for make it shorter
```suggestion
	return c, err
```

pkg/websocket/websocket.go
Atoo35: right this makes sense however i am not so sure how it would work in the codebase. What I mean is that in `gorilla.go` in the `initialize` function, we are sending all the responses to response channel which is `rpc.RPCResponse` type.
If we make a new `Notification` type in rpc we wont be able to write to the channel. 
This would mean we would need to write to another channel which is of the type Notification, eventually adding 2 more functions like the `createReponseChannel` and `getResponseChannel`  for the websocket. Also addition of notificationChannel in the `WebSocket` struct.

Do i understand it correctly?

pkg/gorilla/gorilla.go
```diff
@@ -206,15 +215,29 @@ func (ws *WebSocket) initialize() {
...
+					if !ok {
+						err = fmt.Errorf("unavailable ResponseChannel %+v", res.ID)
+						ws.logger.Logger.Err(err)
+						ws.logger.LogChannel <- err.Error()
+						continue
+					}
+					responseChan <- res
+					close(responseChan)
+				} else {
+					resolved_id := res.Result.(map[string]interface{})["id"]
```
ElecTwix, line 231: we can make it non snake case like linter says `resolvedID`

pkg/websocket/websocket.go
ElecTwix, line 9: Same applies here as well,
people cannot import or use internal packages if we want to use we should move rpc package to pkg

ElecTwix (commented):
Good work
I added some reviews.

db_test.go
Atoo35: yes yes @phughk suggested it, accidentally pushed the commit lol

pkg/gorilla/gorilla.go
phughk: Would the log package not work?

pkg/websocket/websocket.go
phughk: Ah right yes, you are right there - this is relaying protocol to driver. In that case the protocol should be agnostic - this should instead be "Receive async message". That's because in the future, it would be more than one type of message. Addtionally, a web socket doesn't "know" what a live query is at that abstraction level. So you want to avoid mixing layers.

db.go
phughk: I am giving high-level feedback here without digging too much into the implementation.

The driver should not expose internal details at all to users. This should be `Notification` if it is user facing.

ElecTwix: I agree but if we want people to use `rpc.RPCResponse` we should move them to `pkg` folder maybe we should separate the files but I'm not sure if that good idea either.

phughk: We don't want people to use rpc.RPCResponse. This is because, down the line, it might change. And we don't want users to know about that change when and if it does happen.

phughk: I agree that if we did, it would be a separate package though

pkg/gorilla/gorilla.go
Atoo35: its not working. cant see WithField method as well.

timpratim:
Hi @Atoo35, I have added some review comments. The PR is good to be merged once the comments are resolved.

timpratim (changes requested):

timpratim:
They should be visible now. Thanks @Atoo35 .

Atoo35:
yes i can see them now. Will resolve asap. Thanks!

pkg/gorilla/gorilla.go
```diff
@@ -32,7 +32,7 @@ type WebSocket struct {
 	Option   []Option
 	logger   *logger.LogData
 
-	responseChannels     map[string]chan rpc.RPCResponse
+	responseChannels     map[string]chan interface{}
```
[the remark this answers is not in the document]
ElecTwix: Channels are not only used with notifications that it is `interface{}`.

pkg/gorilla/gorilla.go
```diff
@@ -159,15 +168,17 @@ func (ws *WebSocket) Send(method string, params []interface{}) (interface{}, err
 	select {
 	case <-timeout:
 		return nil, ErrTimeout
-	case res := <-responseChan:
-		if res.ID != id {
+	case raw, open := <-responseChan:
+		if !open {
+			return nil, errors.New("channel closed")
+		}
+		res, ok := raw.(rpc.RPCResponse)
```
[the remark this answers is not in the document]
ElecTwix: In this PR we talked into this too but the thing is even if we do the user couldn't use it because the RPC module is internal
if we want to achieve that we need to make rpc package public as well for me we can expose some rpc parts to improve the developer experience.

pkg/gorilla/gorilla.go
```diff
@@ -206,16 +216,38 @@ func (ws *WebSocket) initialize() {
...
+			ws.logger.LogChannel <- err.Error()
+			return
+		}
+		defer close(responseChan)
+		responseChan <- res
+	} else {
+		resolvedID, ok := res.Result.(map[string]interface{})["id"]
+		if !ok {
+			err := fmt.Errorf("response did not contain an 'id' field")
+			ws.logger.Logger.With().Fields(map[string]interface{}{"result": res.Result}).Err(err)
```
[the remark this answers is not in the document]
Atoo35: yes I wasn't getting the desired function here and hence I used the .With().Fields() function

ElecTwix:
I left some comments on some of review's comments.
I think we need the clear them out before implementing.

pkg/gorilla/gorilla.go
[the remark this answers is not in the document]
timpratim: The intention is to allow it to work in embedded mode, and we shouldn't be exposing the implementation details like rpc.

Atoo35:
Thanks @phughk , although this was something I very intially thought needs to be done but I guess we decided in between not to do that if I am not wrong. Anyway thanks for this and sorry for the delay

timpratim:
Hey @ElecTwix, @Atoo35, we added some fixes to the PR today. Do add your review comments regarding them. I will resolve the previous comments after your review comments.

phughk:
@Atoo35 yes, must have misunderstood earlier if you were suggesting that already. Thanks for handling the fixes. Looks like i introduced some linting errors. Would it be possible to fix those by you? Or I can open a PR to do that as well.

timpratim (approved):

phughk:
I have made this an accepted pr for Hacktoberfest @Atoo35 :) Not sure if you are participating

Atoo35:
Ooohhh cool... Yeah it's embarrassing but I forgot about hacktoberfest haha... Will fix the other PR hopefully tonight post work

phughk:
Yeah no rush :)

PR #90, opened 2023-07-31
Refactor SmartUnmarshal

ElecTwix:
SmartUnmarshal output has changed to array
SmartUnmarshal now supports `non-arr normal`, `arr normal` and `arr raw`
fix Typo
fix #79

### DIFF
--- a/db.go
+++ b/db.go
@@ -1,6 +1,7 @@
 package surrealdb
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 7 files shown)


ElecTwix:
Added new code packages for clean up of the code and made modules smaller to consume by other go modules.
Refactor RawUnMarshal to use with RawQuery generic for getting status and other elements that were not possible before.

timpratim:
The changes look good and the code is more readable.

agufagit:
SmartUnmarshal output shouldn't be changed to array, a fair complex project will have most surql code in `BEGIN; ... COMMIT;` block, and it returns exactly what it returns, `RETURN {...}` returns an object, `RETURN true` returns a boolean.  Whether the output should be array depends on user's demands, driver shouldn't do that

ElecTwix:
> SmartUnmarshal output shouldn't be changed to array, a fair complex project will have most surql code in `BEGIN; ... COMMIT;` block, and it returns exactly what it returns, `RETURN {...}` returns an object, `RETURN true` returns a boolean. Whether the output should be array depends on user's demands, driver shouldn't do that

Yeah, you are right, I will look into this but for now we have bigger issues thanks for the feedback I will put this on the list.

agufagit:
> > SmartUnmarshal output shouldn't be changed to array, a fair complex project will have most surql code in `BEGIN; ... COMMIT;` block, and it returns exactly what it returns, `RETURN {...}` returns an object, `RETURN true` returns a boolean. Whether the output should be array depends on user's demands, driver shouldn't do that
> 
> Yeah, you are right, I will look into this but for now we have bigger issues thanks for the feedback I will put this on the list.

I'll contribute and make changes, I kind of need this to work now

Thank you

agufagit:
> > SmartUnmarshal output shouldn't be changed to array, a fair complex project will have most surql code in `BEGIN; ... COMMIT;` block, and it returns exactly what it returns, `RETURN {...}` returns an object, `RETURN true` returns a boolean. Whether the output should be array depends on user's demands, driver shouldn't do that
> 
> Yeah, you are right, I will look into this but for now we have bigger issues thanks for the feedback I will put this on the list.

Never mind, I just see that multiple results can be returned from raw query, your code is the only way to handle that.  I'll just make a function to handle single result case locally in my code.

agufagit:
this code will break in this case though: `[map[result:true status:OK time:4.219606ms]]`, because `RawQuery result` field is set to `[]i` .

also it doesn't handle `throw` error, which is this case: `[map[result:An error occurred: blocked status:ERR time:4.219606ms]]`

PR #91, opened 2023-08-09
Stabilize WS tests

mumoshu:
The tests were failing time to time for me saying this:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9f9f27]

goroutine 23 [running]:
github.com/surrealdb/surrealdb.go/pkg/gorilla.(*WebSocket).initialize.func1()
	/home/mumoshu/p/surrealdb.go/pkg/gorilla/gorilla.go:205 +0xe7
created by github.com/surrealdb/surrealdb.go/pkg/gorilla.(*WebSocket).initialize
	/home/mumoshu/p/surrealdb.go/pkg/gorilla/gorilla.go:196 +0x69
```

The paniic points to https://github.com/surrealdb/surrealdb.go/blob/main/pkg/gorilla/gorilla.go#L205, and I see `ws.logger` being `nil` when the panic happens.

We should make the logger non-nil, or make the `ws.read` goroutine wait until the logger is initialized. Otherwise, every time `ws.read` got an error from the connection (e.g. the connection is closed?), AND the logger is nil (which is always the case for our no-opt `gorilla` ws impl in the tests), the panic happens.

That's said, let's always initialize the logger for now.

We may make the logger getting lazily initialized on demand, but I think that's not a priority for now :)

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -38,15 +38,22 @@ func TestSurrealDBSuite(t *testing.T) {
 	SurrealDBSuite.wsImplementations = make(map[string]websocket.WebSocket)
 
 	// Without options
-	SurrealDBSuite.wsImplementations["gorilla"] = gorilla.Create()
+	logData, err := createLogData(t)
+	require.NoError(t, err)
+	SurrealDBSuite.wsImplementations["gorilla"] = gorilla.Create().Logger(logData)
 
 	// With options
-	buff := bytes.NewBuffer([]byte{})
-	logData, err := logger.New().FromBuffer(buff).Make()
+	logData, err = createLogData(t)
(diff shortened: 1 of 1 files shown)


phughk (approved):
Thanks for fixing and describing!

ElecTwix:
I have identified that there is a bug in the code causing std.out to be used instead of https://github.com/surrealdb/surrealdb.go/blob/main/pkg/logger/logger.go#L51C2-L51C2 when the path is empty. Rest assured, I will create a PR to fix this issue promptly. Thank you for bringing it to my attention.

PR #93, opened 2023-08-09
update version

ElecTwix:
Yesterday, on August 8th, 2023, version 1.21 of Go was released. Since version 1.18 is now outdated, it may be a good idea to update our code base to version 1.20 for improved performance and updated features.


phughk (approved):
Great idea, thank you!

PR #101, opened 2023-09-25
Add Record type for FETCH statement in SELECT.

aoisensi:
I don't have good English so please add comments.

### DIFF
--- a/record.go
+++ b/record.go
@@ -1,0 +1,26 @@
+package surrealdb
+
+import "encoding/json"
+
+type Record[T any] struct {
+	ID     string
+	OnlyID bool
+	Object T
+}
+
+func (r *Record[T]) UnmarshalJSON(data []byte) error {
+	var v any
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+	if v == nil {
+		return nil
+	}
+	if id, ok := v.(string); ok {
+		r.ID = id
+		r.OnlyID = true
+		return nil
+	}
+	r.ID = v.(map[string]any)["id"].(string)
+	return json.Unmarshal(data, &r.Object)
+}
--- a/record_test.go
+++ b/record_test.go
@@ -1,0 +1,74 @@
+package surrealdb_test
+
+import (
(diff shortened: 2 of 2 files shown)


aoisensi:
Maybe `Record.Fetched` is better than `Record.OnlyID`?
Idk.

record.go
```diff
@@ -0,0 +1,26 @@
+package surrealdb
```
ElecTwix, line 1: this should be in in other module not in the main surrealdb package

record_test.go
```diff
@@ -0,0 +1,74 @@
...
+						Bio: "Watch anime Kill Me Baby",
+					},
+				},
+			},
+		},
+	}
+	for _, rule := range rules {
+		t.Run(rule.Name, func(t *testing.T) {
+			var actual testRecordUser
+			err := json.Unmarshal([]byte(rule.Data), &actual)
```
ElecTwix, line 65: shouldn't we also add end to end tests?

record_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"encoding/json"
+	"reflect"
+	"testing"
+
+	"github.com/surrealdb/surrealdb.go"
+)
+
+type testRecordUser struct {
+	ID      string                               `json:"id,omitempty"`
+	Profile surrealdb.Record[*testRecordProfile] `json:"profile,omitempty"`
```
ElecTwix, line 13: why we should have this instead of just basemodel?

ElecTwix (commented):
Thanks for PR :)

I'm not sure why we will need something like this.
users already can unmarshal with the marshal package.
I understand the point of the PR I'm not sure it will help.

Could you give us a beneficial example that this PR has provided?

aoisensi:
I had been using the old version and didn't even know it had been moved to `pkg/marshal`.
Apparently `marshal.Basemodel` is sufficient.
I wish they would update the documentation more!
Thank you.

aoisensi:
Oh, v0.2.1 has too.
Bruh...

PR #108, opened 2023-10-19
added connection interface for non-ws to add in future

ElecTwix:
added connection interface.
connection connect method has been moved to db.New way for simplicity.

### DIFF
--- a/db.go
+++ b/db.go
@@ -5,18 +5,22 @@ import (
 
 	"github.com/surrealdb/surrealdb.go/pkg/model"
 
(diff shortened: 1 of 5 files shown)


ElecTwix:
Hi @timpratim and @phughk, 
This pull request has been open for about a month. Could you please check it when you have a moment?

This PR is also a blocker for me cuz I would like to implement HTTP for the go driver and WS connection is not the right term.

phughk (approved):
lgtm, thanks!

PR #114, opened 2023-11-22
move mapToStruct to marshal package

### LINKED ISSUE #106: Feature: Move and test `unmarshalMapToStruct(data map[string]interface{}, outStruct interface{})`

A lot of the deserialisation is happening via JSON, even for code that is already deserialised.
There is a non-json marshaller, `unmarshalMapToStruct(data map[string]interface{}, outStruct interface{})`, in gorilla.go.
It should be moved to a separate package (not protocol implementation related, not API related) and tested.

### Describe the solution

Move the function `unmarshalMapToStruct(data map[string]interface{}, outStruct interface{})` into a separate package.
Test the funciton independently of protocol implementations.

### Alternative methods

There could already be code that does this. It's important that we don't rely on translating `map[string]interface{}` into json and back into a golang struct though.

### SurrealDB version

1.0.0

### Contact Details

pratim.bhosale@surrealdb.com hugh@surrealdb.com

### Code of Conduct

ElecTwix:
added unit tests
fix #106

### DIFF
--- a/pkg/marshal/marshal.go
+++ b/pkg/marshal/marshal.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strconv"
 
 	"github.com/surrealdb/surrealdb.go/pkg/constants"
 	"github.com/surrealdb/surrealdb.go/pkg/util"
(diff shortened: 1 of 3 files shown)


pkg/marshal/marshal.go
```diff
@@ -162,3 +163,77 @@ func SmartMarshal[I any](inputfunc interface{}, data I) (output interface{}, err
 	}
 	return nil, ErrNotValidFunc
 }
+
+// UnmarshalMapToStruct unmarshals a map[string]interface{} to a struct.
+func UnmarshalMapToStruct(data map[string]interface{}, outStruct interface{}) error {
```
phughk, line 168: I'm ok with this moving, but we need to remember that this will change with the introduction of Column and Rows

phughk (approved):
lgtm, but noting that this code is likely to change down the line as we introduce Row and Column.
Will merge after resolution

pkg/marshal/marshal.go
ElecTwix: Yes, of course, but for now, we should continue to improve the project in meanwhile.

tobiemh:
Hi @ElecTwix thanks so much for submitting this pull request 😃 👍 - it's greatly appreciated, even after so much time has passed! Since the date of this PR, the [Golang SDK](https://github.com/surrealdb/surrealdb.go) has seen significant changes, with support for both HTTP and WebSocket connections, full support of the SurrealDB RPC methods, a binary communication protocol using CBOR, and support for embedding SurrealDB in memory and with SurrealKV coming soon. As a result, I'm going to close this PR as it conflicts with the current codebase which has diverged significantly. Feel free to add any feedback, issues, or further pull requests to the new SDK code 🚀 🎉 !

PR #115, opened 2023-11-29
build: Add tests for fetch using passing arguments

### LINKED ISSUE #40: Getting Parsing Error When Use Fetch with map[string]interface
Hi Everyone,

I was selecting the record and fetching it is the field record link with the fetch keyword
but when I try to use it with the parameter it gives
 `There was a problem with the database: Parse error when parsing 'fetch $fetchstr'`
this error I checkout the source code [this line](https://github.com/surrealdb/surrealdb.go/blob/main/internal/websocket/ws.go#L112) JSON Marshaling giving the error.
Also, I tried with not safe method string merge ( fmt.sprintf ) and added my string at the end it worked that way.

Am I doing something wrong?
if that's the case please let me know.

The Code I'm trying to fetch

```go
res, err := db.Globaldb.Query("select * from $table fetch $fetchstr;", map[string]interface{}{
	"table":  ID,
	"fetchstr": fetchstr,
})

```

## UPDATE:
for clear out
### The problem I was facing was:
After the FETCH statement, arguments are no longer parsed.
### Example:
`select * from $table FETCH $fetchstr;`
$**table** will parse but $**fetchstr** will not parse

sbshah97:
Added 5 different test cases to test the fetch functionality. 
The remaining tests need to be changed once the functionality is fixed at the root source

Closes #40

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -121,10 +121,10 @@ func (s *SurrealDBTestSuite) SetupSuite() {
 // Can be used with any user
 func signin(s *SurrealDBTestSuite) interface{} {
 	authData := &surrealdb.Auth{
-		Database:  "test",
-		Namespace: "test",
-		Username:  "root",
-		Password:  "root",
+		// Database:  "test",
+		// Namespace: "test",
+		Username: "root",
+		Password: "root",
 	}
 	signin, err := s.db.Signin(authData)
 	s.Require().NoError(err)
--- a/db_test.go
+++ b/db_test.go
@@ -36,10 +36,18 @@ type SurrealDBTestSuite struct {
 // a simple user struct for testing
 type testUser struct {
 	marshal.Basemodel `table:"test"`
-	Username          string   `json:"username,omitempty"`
-	Password          string   `json:"password,omitempty"`
-	Friends           []string `json:"friends,omitempty"`
-	ID                string   `json:"id,omitempty"`
+	Username          string `json:"username,omitempty"`
+	Password          string `json:"password,omitempty"`
+	ID                string `json:"id,omitempty"`
+}
+
+// a simple user struct for testing
+type testUserWithFriend[I any] struct {
+	marshal.Basemodel `table:"user_with_friend"`
+	Username          string `json:"username,omitempty"`
+	Password          string `json:"password,omitempty"`
+	ID                string `json:"id,omitempty"`
+	Friends           []I    `json:"friends,omitempty"`
 }
 
 func TestSurrealDBSuite(t *testing.T) {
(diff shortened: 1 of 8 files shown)


db_test.go
```diff
@@ -226,6 +225,78 @@ func (s *SurrealDBTestSuite) TestDelete() {
...
+			s.Error(err)
+			s.Nil(res)
+		}
+	})
+
+	s.Run("Run fetch on hardcoded query", func() {
+		query := "SELECT * from users fetch friends.*"
+		res, err := s.db.Query(query, map[string]interface{}{})
+		s.NoError(err)
+		s.NotNil(res)
```
phughk, line 269: We probably want to verify the output of the query to make sure it works.

db_test.go
```diff
@@ -226,6 +225,78 @@ func (s *SurrealDBTestSuite) TestDelete() {
...
+
+	// User rows are individually fetched
+	s.Run("Run fetch for individual users", func() {
+		for _, v := range userSlice {
+			res, err := s.db.Query("select * from $table fetch $fetchstr;", map[string]interface{}{
+				"record":   v.ID,
+				"fetchstr": "friends.*",
+			})
+			// TODO: This should be fixed once the code is fixed
+			s.Error(err)
```
phughk, line 260: Could you please change this from `Error` to `NoError`, use `s.Skip` at the start of the test, and create a ticket that this needs to be resolved please? 🙏
```suggestion
			s.NoError(err)
```

db_test.go
```diff
@@ -226,6 +225,78 @@ func (s *SurrealDBTestSuite) TestDelete() {
...
+			s.Nil(res)
```
phughk, line 261: This can also be s.NotEmpty for now

db_test.go
```diff
@@ -226,6 +225,78 @@ func (s *SurrealDBTestSuite) TestDelete() {
...
+		s.NoError(err)
+		s.NotNil(res)
+	})
+
+	s.Run("Run fetch on query using map[string]interface{} for thing and fetchString", func() {
+		res, err := s.db.Query("select * from $record fetch $fetchstr;", map[string]interface{}{
+			"record":   "users",
+			"fetchstr": "friends.*",
+		})
+		// TODO: This should be fixed once the code is fixed
```
phughk, line 277: Also flip logic, skip, and ticket please

db_test.go
```diff
@@ -226,6 +225,78 @@ func (s *SurrealDBTestSuite) TestDelete() {
...
+		s.Error(err)
+		s.Nil(res)
+	})
+
+	s.Run("Run fetch on query using map[string]interface{} for fetchString", func() {
+		res, err := s.db.Query("select * from users fetch $fetchstr;", map[string]interface{}{
+			"fetchstr": "friends.*",
+		})
+		// TODO: This should be fixed once the code is fixed
```
phughk, line 286: Same here

db_test.go
```diff
@@ -226,6 +225,78 @@ func (s *SurrealDBTestSuite) TestDelete() {
...
+		s.Error(err)
+		s.Nil(res)
+	})
+
+	s.Run("Run fetch on query using map[string]interface{} for thing or tableName", func() {
+		res, err := s.db.Query("select * from $record fetch friends.*;", map[string]interface{}{
+			"record": "users",
+		})
+		s.NoError(err)
+		s.NotNil(res)
```
phughk, line 296: Could the output be verified here please?

phughk (commented):
Great work, thank you for doing this! Left some comments for things to improve.

ElecTwix:
Maybe we can use `unmarshalRaw` or `smartUnmarshal` to make it simpler because there is a lot of casting.

sbshah97:
> Maybe we can use unmarshalRaw or smartUnmarshal to make it simpler because there is a lot of casting.

Yep I was thinking of doing it in the follow up PR once this goes in. 
Can you help me with how those would be used here?

I am not sure how to use those methods, at the end of the README there were some examples but didn't really help.

phughk:
We should wait for the unmarshal improvements before merging this. Appreciate the patience and diligence 👍

ElecTwix:
opened a PR to @sbshah97 's repo https://github.com/sbshah97/surrealdb.go/pull/1 for unmarshal.

db_test.go
```diff
@@ -226,6 +225,129 @@ func (s *SurrealDBTestSuite) TestDelete() {
 	s.Require().NoError(err)
 }
 
+// TODO: This can maybe be replaced by smartUnmarshal
+func (s *SurrealDBTestSuite) validateFriends(actualUser interface{}, expectedUserID string) {
+	s.T().Helper()
+	output, ok := actualUser.([]interface{})
+	if !ok {
```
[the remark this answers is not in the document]
sbshah97: This is not working for some reason, using assert.True instead

sbshah97:
@phughk could you review this again?

PR #120, opened 2024-01-01
feat(marshal): smartUnmarshalRaw1 to handle raw query single return result including THROW statement errors

agufagit:
current SmartUnmarshal only handles array return results, doesn't handle other return types such as `create only user` which returns an object.

the new SmartUnmarshalRaw1 only handles raw query, and only handles single return result: [map[result:true status:OK time:4.219606ms]]

This is required because of custom functions `DEFINE FUNCTION fn::xxx` and `BEGIN;...COMMIT;` block doesn't return arrays, and only returns single result.

it also handles `THROW` statement errors `BEGIN; THROW 'blocked'; COMMIT;` : [map[result:An error occurred: blocked status:ERR time:4.219606ms]]

this function should cover 99% of people's needs

tested against surrealdb v1.0.0

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -228,7 +228,7 @@ func (s *SurrealDBTestSuite) TestDelete() {
 
 func (s *SurrealDBTestSuite) TestInsert() {
 	s.Run("raw map works", func() {
-		userData, err := s.db.Insert("user", map[string]interface{}{
+		userData, err := s.db.Insert("users", map[string]interface{}{
 			"username": "johnny",
 			"password": "123",
 		})
@@ -244,7 +244,7 @@ func (s *SurrealDBTestSuite) TestInsert() {
 	})
 
 	s.Run("Single insert works", func() {
-		userData, err := s.db.Insert("user", testUser{
+		userData, err := s.db.Insert("users", testUser{
 			Username: "johnny",
 			Password: "123",
 		})
@@ -268,7 +268,7 @@ func (s *SurrealDBTestSuite) TestInsert() {
 			Username: "johnny2",
 			Password: "123",
 		})
-		userData, err := s.db.Insert("user", userInsert)
+		userData, err := s.db.Insert("users", userInsert)
 		s.Require().NoError(err)
 
 		// unmarshal the data into a user struct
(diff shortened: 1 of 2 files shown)


ElecTwix:
Thanks for the PR @agufagit,
I was working on my project and I faced similar problems with current driver.
It can be good to add something like this as a helper function.

@phughk , @timpratim what do you think?

db_test.go
```diff
@@ -228,7 +228,7 @@ func (s *SurrealDBTestSuite) TestDelete() {
 
 func (s *SurrealDBTestSuite) TestInsert() {
 	s.Run("raw map works", func() {
-		userData, err := s.db.Insert("user", map[string]interface{}{
+		userData, err := s.db.Insert("users", map[string]interface{}{
```
ElecTwix, line 231: If we change these ones
please add them to the commit message

db_test.go
```diff
@@ -244,7 +244,7 @@ func (s *SurrealDBTestSuite) TestInsert() {
 	})
 
 	s.Run("Single insert works", func() {
-		userData, err := s.db.Insert("user", testUser{
+		userData, err := s.db.Insert("users", testUser{
```
ElecTwix, line 247: same here

db_test.go
```diff
@@ -268,7 +268,7 @@ func (s *SurrealDBTestSuite) TestInsert() {
 			Username: "johnny2",
 			Password: "123",
 		})
-		userData, err := s.db.Insert("user", userInsert)
+		userData, err := s.db.Insert("users", userInsert)
```
ElecTwix, line 271: same here

db_test.go
```diff
@@ -574,6 +574,47 @@ func (s *SurrealDBTestSuite) TestSmartUnMarshalQuery() {
...
 }
 
+func (s *SurrealDBTestSuite) TestSmartUnMarshalRaw1Query() {
+	user := testUser{
+		Username: "electwix",
+		Password: "1234",
+	}
+
+	s.Run("raw create query", func() {
+		QueryStr := "Create Only users set Username = $user, Password = $pass"
```
ElecTwix, line 584: maybe we can use full uppercase for SurrealQL operators and statement such as CREATE, ONLY

pkg/marshal/marshal.go
```diff
@@ -21,6 +21,13 @@ type RawQuery[I any] struct {
 	Detail string `json:"detail"`
 }
 
+type RawQueryRaw[I any] struct {
```
ElecTwix, line 24: Maybe we can call it RawQuerySingle or something along those lines or we can change the original RawQuery to me indeed of an array of I and then refactor according to that

ElecTwix (changes requested):
Looks good overall,
did some nitpicks.

pkg/marshal/marshal.go
agufagit: original RawQuery only handles when `result` field is array, but `result` field can be an object or basic types, golang doesn't allow `any` to be casted to `[]any` and vice versa, that's why a new struct and a new funciton are created.

ElecTwix: Yeah, you're right my bad.

phughk (approved):
We are very likely to change direction with the API. Until then, approving this, but please consider that it might not live long as available functionality.

PR #121, opened 2024-01-09
Add context.Context support to the API

### LINKED ISSUE #100: Feature: context.Context Support

This would be a part of a wider change to allow trace information propagation (if possible). Passing in the context is currently not possible and we would need it to propagate the trace information. Seeing how the Surreal server already supports tracing, this would provide a very good method of troubleshooting entire requests.

### Describe the solution

The standard way to do so is by using `context.Context` and passing it as the first argument to the function. eg.

```go
func (db *DB) Create(ctx context.Context, thing string, data any) (any, error)
```

This would be the preferred way as the SDK has not yet reached v1.0 so it's easier to change the method signature without a v2 release.

### Alternative methods

If the method signatures are not to be changed, the `Context` suffix can be added to each method. This is rather verbose to type and is typically used when a pre-existing API is adding context support.

eg.

```go
func (db *DB) CreateContext(ctx context.Context, thing string, data any) (any, error)
```

There is no other obvious way to add context information.

### SurrealDB version

N/A

### Contact Details

_No response_

### Code of Conduct

VoyTechnology:
The change adds context.Context support to the APIs. The context is
currently unused and just serves as an API change before more
functionality is added (eg. tracing).

This is a breaking API change.

Closes #100

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ type User struct {
 
 func main() {
 	// Connect to SurrealDB
-	db, err := surrealdb.New("ws://localhost:8000/rpc")
+	db, err := surrealdb.New(context.Background(), "ws://localhost:8000/rpc")
 	if err != nil {
 		panic(err)
 	}
@@ -47,11 +47,11 @@ func main() {
 		Username:  "root",
 		Password:  "root",
 	}
-	if _, err = db.Signin(authData); err != nil {
+	if _, err = db.Signin(context.Background(), authData); err != nil {
 		panic(err)
 	}
 
-	if _, err = db.Use("test", "test"); err != nil {
+	if _, err = db.Use(context.Background(), "test", "test"); err != nil {
 		panic(err)
 	}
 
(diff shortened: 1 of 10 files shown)


ElecTwix:
Hi @VoyTechnology,

Thanks for your PR.

I understand the importance of using contexts, but I'm not sure we should add them while we're not utilizing them in a useful way yet.

We should also consider that we'll be making some major changes in the future to try to be compatible with the golang standard interface ([database/sql](https://pkg.go.dev/database/sql)). Once these changes are implemented, I think we could then consider adding contexts.


Please let me know what you think. We can discuss this further if you'd like.

tobiemh:
Hi @VoyTechnology thanks so much for submitting this pull request 😃 👍 - it's greatly appreciated, even after so much time has passed! Since the date of this PR, the [Golang SDK](https://github.com/surrealdb/surrealdb.go) has seen significant changes, with support for both HTTP and WebSocket connections, full support of the SurrealDB RPC methods, a binary communication protocol using CBOR, and support for embedding SurrealDB in memory and with SurrealKV coming soon. As a result, I'm going to close this PR as it conflicts with the current codebase which has diverged significantly. Feel free for any feedback, issues, or further pull requests to the new SDK code 🚀 🎉 !

PR #131, opened 2024-03-28
Feature: Added support for multiple entries

kharbanda14:
Added db method to insert multiple records at once, with updated readme file and test.

Method `CreateMany` calls `Query` method with the insert query based on the documentation.

Surreal DB Version : 1.3.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,14 @@ func main() {
 		panic(err)
 	}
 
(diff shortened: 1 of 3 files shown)


kharbanda14:
After searching, the bulk insert functionality already exists by using `db.Insert()` method.

PR #135, opened 2024-04-28
Panic fix on gorilla ws when connection lost.

ElecTwix:
fix linter errors with shadow checking
upgrade linter

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,16 +12,16 @@ permissions:
 jobs:
   golangci:
     permissions:
-      contents: read  # for actions/checkout to fetch code
-      pull-requests: read  # for golangci/golangci-lint-action to fetch pull requests
+      contents: read # for actions/checkout to fetch code
+      pull-requests: read # for golangci/golangci-lint-action to fetch pull requests
     name: lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v5
         with:
-          go-version: '>=1.18'
-      - uses: actions/checkout@v3
+          go-version: ">=1.18"
+      - uses: actions/checkout@v4
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
+        uses: golangci/golangci-lint-action@v5
         with:
           args: --timeout=5m
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -34,15 +34,16 @@ linters-settings:
       - condition
       - return
     ignored-numbers:
-      - '0'
-      - '1'
-      - '2'
-      - '3'
+      - "0"
+      - "1"
+      - "2"
+      - "3"
     ignored-functions:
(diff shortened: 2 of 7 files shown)


ElecTwix:
To be Noted: this PR changes how to create a connection it added a new step to call `db.Initialize` method to get ws start reading from db.
this is needed because if something goes bad with the read it needs to return an error instead of panicking and crashing the program.

.github/workflows/lint.yml
```diff
@@ -12,16 +12,16 @@ permissions:
...
     permissions:
-      contents: read  # for actions/checkout to fetch code
-      pull-requests: read  # for golangci/golangci-lint-action to fetch pull requests
+      contents: read # for actions/checkout to fetch code
+      pull-requests: read # for golangci/golangci-lint-action to fetch pull requests
     name: lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v5
```
phughk, line 20: Bit of a nitpick here, but will happily glance over it for this PR - could such changes be included as separate PRs? It makes them easier to review and diagnose when something goes wrong.

phughk (commented):
Thanks! Looks good, but I am wondering about the need for a separate step. That step will always be immediately called after Connect. Is it possible to tie it together with that?

ElecTwix:
> Thanks! Looks good, but I am wondering about the need for a separate step. That step will always be immediately called after Connect. Is it possible to tie it together with that?

I tried that first but I'm not sure it is possible, we can add an error channel but I'm not sure that it will be easily understandable I think this is more straightforward than that.
Still, I will experiment and get back if I can manage to create something like that.

.github/workflows/lint.yml
ElecTwix: > Bit of a nitpick here, but will happily glance over it for this PR - could such changes be included as separate PRs? It makes them easier to review and diagnose when something goes wrong.

opened #136 for separate PR.

ElecTwix:
Hi @phughk,

After some testing, I've added a channel-based version of the implementation. I'm not sure if it's an improvement, and I hesitated to include it here since it might not be accepted. However, here's the commit for reference: https://github.com/ElecTwix/surrealdb.go/commit/908a377794436715b35dac3dd7ff2af488c4f5af

tobiemh:
Hi @ElecTwix thanks so much for submitting this pull request 😃 👍 - it's greatly appreciated, even after so much time has passed! Since the date of this PR, the [Golang SDK](https://github.com/surrealdb/surrealdb.go) has seen significant changes, with support for both HTTP and WebSocket connections, full support of the SurrealDB RPC methods, a binary communication protocol using CBOR, and support for embedding SurrealDB in memory and with SurrealKV coming soon. As a result, I'm going to close this PR as it conflicts with the current codebase which has diverged significantly. Feel free to add any feedback, issues, or further pull requests to the new SDK code 🚀 🎉 !

PR #138, opened 2024-05-21
fix #119, panic gorilla ws when connection lost.

### LINKED ISSUE #119: Bug: Unhandled errors in websockets can lead to panics
### Describe the bug

https://github.com/surrealdb/surrealdb.go/pull/118/files

Not much detail, but there is a code line. Apparently, when the error isnt handled correctly, the websocket connections can panic.

The TODO is going to be added to the codebase

### Steps to reproduce

Unknown, may need to review code

### Expected behaviour

No panics, errors handled correctly

### SurrealDB version

1.0.0+20231130.a420c821.dirty for macos on aarch64

### Contact Details

hugh@surrealdb.com

### Code of Conduct

ElecTwix:
instead of panic it will return error when any send function call when connection lost.

### DIFF
--- a/pkg/conn/gorilla/gorilla.go
+++ b/pkg/conn/gorilla/gorilla.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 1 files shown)


ElecTwix:
TLDR: This pull request aims to address issue #135 without requiring an additional function call.

timpratim:
Hi @ElecTwix , thank you for the PR. Everything looks great. It would be great if we could add a test case for the `Send` method in the same PR. Happy to merge the PR then.

pkg/conn/gorilla/gorilla.go
```diff
@@ -235,25 +243,38 @@ func (ws *WebSocket) write(v interface{}) error {
...
-					continue
+	for {
+		select {
+		case <-ws.closeChan:
+			return
+		default:
+			var res rpc.RPCResponse
+			err := ws.read(&res)
+			if err != nil {
+				shouldExit := ws.handleError(err)
```
hulkingshtick, line 254: The function should exit on **any** error returned from `ws.Conn.ReadMessage`.   You can fix by inlining `WebSocket.read` here and exiting when the call to `ws.Conn.ReadMessage` returns an error.  Because `handleError` does not accurately detect all websocket errors, this thing can loop endlessly and eventually panic (the gorilla detects useless application loops).

PR #142, opened 2024-06-13
feature-141-upsert-command

### LINKED ISSUE #141: Feature: UPSERT

The new (`v2.0.0-alpha.1`) `UPSERT` command is not available in the Go SDK.

### Describe the solution

An implementation of the `UPSERT` command. 

### Alternative methods

I could not find any way to make the `UPSERT` command work without an "official" implementation, since the `send` function is private. 

### SurrealDB version

2.0.0-alpha.1 for linux on x86_64

### Contact Details

_No response_

### Code of Conduct

EtienneBruines:
Add support for the new UPSERT command. Fixes #141


ElecTwix:
Hi @EtienneBruines, thanks for your PR.
PR looks good but I think we should wait Release 2.0.0 to came out because it is `pre-release` at the moment.

EtienneBruines:
Since there is no workaround available to use the new `UPSERT`, perhaps an [`alpha` release](https://go.dev/doc/modules/version-numbers) makes sense? :thinking:

lawrencejob:
Hi all, since 2.0.0 is live, it would be really valuable to have this feature. I updated my DB and now can't use it, unless I'm missing something obvious

PR #160, opened 2024-10-17
Additional CBOR types

remade:
- Add Custom types with CBOR marshaling/unmarshaling
- Test and fix CBOR inconsistencies
- Add test

### DIFF
--- a/db.go
+++ b/db.go
@@ -199,7 +199,7 @@ func Query[TResult any](db *DB, sql string, vars map[string]interface{}) (*[]Que
 	return res.Result, nil
 }
 
-func Create[TResult any, TWhat models.TableOrRecord](db *DB, what TWhat, data interface{}) (*TResult, error) {
+func Create[TResult any, TWhat TableOrRecord](db *DB, what TWhat, data interface{}) (*TResult, error) {
 	var res connection.RPCResponse[TResult]
 	if err := db.con.Send(&res, "create", what, data); err != nil {
 		return nil, err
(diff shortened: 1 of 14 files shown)


agufagit:
Hi,

the returning result is `&[{Status:OK Time:28.641008ms Result:map[id:{Table:user ID:ls6bchgtj8l721a8dyeu}]}]`,
can you make returning result `&[{Status:OK Time:28.641008ms Result:map[id:user:ls6bchgtj8l721a8dyeu]}]`?  Which is compatible with old driver, and I don't see the use to make an ID an object, it's supposed to be a string.

Also I'm getting another error `"returnRawError":"json: unsupported type: map[interface {}]interface {}"`.

For `Query` function, the return type doesn't need to be a pointer, it can be `[]QueryResult[TResult]`, golang is different from rust, slices already contain a pointer, and are designed to be passed "as is" for data locality

remade:
Hi @agufagit 

With CBOR implemented, all responses are CBOR encoded including the record ID. This is to our advantage. As @kearfy mentioned, there is a `String()` method on the RecordID type you can use to retrieve the familiar representation. Also, there are helper methods to create/parse record IDs and other custom types. As part of this PR, I will make sure all types have `String()` and `SurrealString()` methods

agufagit:
Hi,

I hope this doesn't mean I have to loop every record to call id.String(), dateTime.toTime()? 

Suppose I have a struct

```go
type Data struct {
  ID string `json:"id"`
  Date time.Time `json:"date"`
}
```

will driver be smart enough to automatically convert id to string, date to time.Time? Or we have to do it ourselves?

remade:
@agufagit 

Whenever you are retrieving the ID, when you retrieve the ID, you can simply call the string method
```
d.ID // for RecordID
d.ID.String() // for string representation
```

agufagit:
that's what I want to avoid to do, my app is fairly complex, that would mean too much work.  I'm using auto gen graphql, where ID is string, I'll have to redefine every struct and map it to graphql struct instead of just using auto generated graphql struct

but no biggie, once you are done, I'll just fork the repo and make changes, and use that forked repo instead

remade:
@agufagit 
Have you tried it out? When a string value is expected of any type, as far as it implements the `Stringer` interface, all should be well. You don't need to do anything extra. If you experience otherwise, it definitely worth resolving

agufagit:
I didn't try the new commit yet, but that sounds great, I'll try it after you're done and new release is out for v2.0.0, thank you.

PR #168, opened 2024-10-29
Handle RPC errors

### LINKED ISSUE #172: Bug: RPC errors logged, not gracefully handled
### Describe the bug

When a query fails, such as invalid syntax, an RPC error is returned. However, at the moment this is only logged to the terminal rather than returned as an error meaning there is no way to tell if a query failed on the database.

### Steps to reproduce

Query the database with a query using incorrect syntax. The database will reject this and send back an RPC error logged in the terminal.

### Expected behaviour

Any errors the database encounters should be returned to `Send()` so it can be returned and handled gracefully.

### SurrealDB version

surreal 2.0.4

### Contact Details

_No response_

### Code of Conduct

TheRangiCrew:
When a query fails, such as invalid syntax, an RPC error is returned. However, at the moment this is only logged to the terminal rather than returned as an error meaning there is no way to tell if a query failed on the database.

Wondering if we can fix this...

Fixes #172 

The idea is to just create another response channel in the WebSocket that listens for errors. The `Send()` method could then use that channel to return an `error` which would be the message returned from SurrealDB. Previously an error would never be returned by `Send()` leaving any failed queries to still wait for a response, preventing higher functions from returning.

### DIFF
--- a/db.go
+++ b/db.go
@@ -197,9 +197,6 @@ func Query[TResult any](db *DB, sql string, vars map[string]interface{}) (*[]Que
 	if err := db.con.Send(&res, "query", sql, vars); err != nil {
 		return nil, err
 	}
-	if res.Error != nil {
-		return nil, res.Error
-	}
 
 	return res.Result, nil
 }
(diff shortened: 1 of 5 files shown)


TheRangiCrew:
Fixes #172

remade (approved):

PR #199, opened 2024-12-24
Added built libs

remade:


clarkmcc:
Looking forward to getting embedded mode back in! Couple questions:
- It looks like there's only Windows/Linux support here, what's the probability of getting macOS support as well?
- Does it make sense to write a script that pulls the source and builds each shared library for each OS so that it's easier to keep them up to date?
- Does it make sense to check these in with git LFS?

Issue #207, opened 2025-07-10
Bug: Json decode to map[interface]interface and not map[string]interface


MaxThom:
### Describe the bug

Hi,
Im upgrading to latest db and sdk version. Everything is working with v1, but I think the error comes from CBOR.

I have the following struct that is stored in the db:
```go
type DeviceProperties struct {
	Desired  map[string]interface{} `json:"desired,omitempty" yaml:"desired"`
	Reported map[string]interface{} `json:"reported,omitempty" yaml:"reported"`
}
```
Which result in this in the database:
```json
properties: {
		desired: {
			"swarm.v1.DataRateProp": {
				sec: 5
			}
		}
	},
```
but when I read my struct, the Desired and Reported field value interface{} are map[interface]interface instead of map[string]interface which was not the case in v1. This create issues as the json marshaller cannot marshal this type.

Is there a way to cast this back to map[string]any instead of map[any]any? json can only be map[string]any anyway! Thnk you

Look like It is a CBOR thing and I found a possible solution: https://github.com/fxamacker/cbor/issues/654

I did a fork where I changed the cbor decoder in the file pkgs/model/cbor.go to:

```go
func getCborDecoder() cbor.DecMode {
	tags := registerCborTags()
	dm, err := cbor.DecOptions{
		TimeTagToAny:   cbor.TimeTagToTime,
		DefaultMapType: reflect.TypeOf(map[string]any(nil)),
	}.DecModeWithTags(tags)
	if err != nil {
		panic(err)
	}

	return dm
}
``` 

*added `DefaultMapType: reflect.TypeOf(map[string]any(nil))`*

This solve all my issues on marshalling the saved data to JSON. I tried a PR but can't push to a branch.
And maybe there is a more proper way?

@tobiemh 

### Steps to reproduce

Create a field map[string]any, populate it with submaps and write to db.
Read the struct and marshal the submap.

### Expected behaviour

the submap should be of type map[string]any and not map[any]any

### SurrealDB version

2.6.2

### Contact Details

maxthomassin@hotmail.com

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mumoshu:
Hi @MaxThom! Thank you very much for the detailed report, and your pull request!
I've merged #208 and cut v0.3.3 just now.
Could you confirm it did fix the issue for your specific use-case?

Thank you in advance for your help!

[...]

PR #208, opened 2025-07-10
Added specific map type to map[string]any for JSON

MaxThom:
CBOR decode map to map[any]any which causes the JSON marshaller to fail as it is not a supported type. We can force the map type to map[string]any in the CBOR decoder.

See issue #207 
Found solution: [CBOR](https://github.com/fxamacker/cbor/issues/654)

Questionning:
1. Does surreal support different field key than string?
2. Should we open the decoder options as parameter?

### DIFF
--- a/pkg/models/cbor.go
+++ b/pkg/models/cbor.go
@@ -123,7 +123,8 @@ func getCborEncoder() cbor.EncMode {
 func getCborDecoder() cbor.DecMode {
 	tags := registerCborTags()
 	dm, err := cbor.DecOptions{
-		TimeTagToAny: cbor.TimeTagToTime,
+		TimeTagToAny:   cbor.TimeTagToTime,
+		DefaultMapType: reflect.TypeOf(map[string]any(nil)),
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
Hi @MaxThom! Thank you very much for reporting the issue and submitting this pull request!
This LGTM- I've also added a regression test that complements your pull request, so that we never encounter this again.

Re: your question, I believe SurrealDB does not support number-like field names within objects, so no toggle sounds good.

You can double-check this behavior by seeing how number-like keys are turned into strings:

```
ns1/db1> create person SET f1 = object::from_entries([[1,1]])
[[{ f1: { "1": 1 }, id: person:h8r0tnjsmhlk3jt5zp81 }]]

ns1/db1> create person SET f1 = {1:2}
[[{ f1: { "1": 2 }, id: person:cl5xczuem4al9atmnj22 }]]

ns1/db1> create person SET f1 = {"1":3}
[[{ f1: { "1": 3 }, id: person:kceossao0y9ea7eq4u77 }]]
```

Issue #224, opened 2025-07-15
Documentation: How to do a Bulk Insert via the Query rpc method


nickchomey:
### Description

I would like to do a Bulk Insert via the Query method, so that I can use `RETURN NONE`, which isn't possible with the Insert method, as per #203. 

I see [this example](https://github.com/surrealdb/surrealdb.go/blob/c8b2e1584aea3b737ab8547e0623b7308f156c01/example/query_test.go#L32-L47) for creating a single record. 

I tried something similar, but nothing was getting written to the db.

The only way I can get bulk insert to work via Query is to literally just marshal the records and then convert to a string [like this](https://github.com/nickchomey/conduit-connector-surrealdb/blob/8fbfd3444fbc738796564752359ae9869d347042/destination/destination.go#L331-L351). But there's a lot of extra overhead like that. 

I'm hoping that the issue here is just that I'm not aware of how to structure the query string and param to handle bulk payload. 

Or is this just a limitation inherent to the [RPC Protocol](https://surrealdb.com/docs/surrealdb/integration/rpc#query)?

 

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #202, opened 2025-03-19
Bug: Responses seem to be unmarshalled twice


nickchomey:
### Describe the bug

Ive been working on a CDC tool for populating Surreal from other datasources. I just imported 1 million rows from mysql and took a pprof profile of it. I noticed that there's two nearly-identical and very large sections in the flamegraph related to CborUnmarshaler

![Image](https://github.com/user-attachments/assets/bcd22f2c-6269-4697-9567-dda5388bf6b4)

One is from `(*WebSocketConnection).Send` and the other is from  `(*WebSocketConnection).handleResponse` (which unmarshals multiple times...?)

Surely this is not necessary?

### Steps to reproduce

I used bulk insert, but surely this happens with any response. just take a pprof profile, or step debug. 

### Expected behaviour

Only unmarshal the response once

### SurrealDB version

server 2.2.1, go sdk 0.3.2 

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mumoshu:
@nickchomey Thank you very much for your detailed report and feedback. Much appreciated!
That does look problematic! Let me see how we can fix it. Probably I'll start by trying to extract the cbor decoder into a field within the marshaller, so that you could reuse it for the lifetime of the marshaller.

mumoshu:
Hi @nickchomey! Thank you again for your detailed feedback!

Could you give #222 a try on your environment?
It should remove all the unnecessary instantiations of the custom CBOR encoders and decoders in the hot path, without breaking anything.

mumoshu:
Thank you @nickchomey!
It isn't really obvious how we do it due to how Go, WebSocket, and SurrealDB RPC responses work- but here's my answer: #231

mumoshu:
@nickchomey Thank you for your confirmation!

Maybe my explanation was not clear enough!
We need to further improve the underlying CBOR library to achieve even greater efficiency.
From what I can see from the flame graphs and the simple arithmetic, we managed to make it the best it can be today, thanks to your feedback 👍 

Yeah, let's close this as resolved and move on to other things!

FYI, I'm planning to work on these milestones: https://github.com/surrealdb/surrealdb.go/milestones. Please let me know if you have anything more important.

We are [on Discord](https://discordapp.com/channels/902568124350599239/1013528025045938207) too, so come and say hi if you don't mind 👋 

Thank you again for all your support @nickchomey!

PR #227, opened 2025-07-16
Add testable example for bulk inserts and upserts

mumoshu:
Ref #224

### DIFF
--- a/example/connection.go
+++ b/example/connection.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 
 	surrealdb "github.com/surrealdb/surrealdb.go"
-	"github.com/surrealdb/surrealdb.go/pkg/models"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


mumoshu:
Hi @nickchomey! Let me reply one by one.

> can we just send field params as strings and the db will handle the appropriately?

It depends on where those strings are, but generally speaking, no- you should leave placeholders `$varname` in the query, while passing query vars as `map` whose values are of the types that match the field types in SurrealDB.

> we can't define structs even with string types in such a situation...

Yeah same happened to me when building the connector!

> How do you handle schemaful tables and fields in the fivetran connector?

I think I've done the same. I defined a schemaful table just to be sure, although I could have used a schemaless one.

https://github.com/surrealdb/fivetran-destination/blob/7e665bdd2aa011dfdc15dc04aa92e39b06b61982/internal/connector/table_define.go#L22

> Perhaps there's an issue with bulk insert via Insert and Query that the sdk can fix/handle behind the scenes?

Interesting! If I understand correctly, you discovered that batch inserting arrays of structs using `insert` or `query` stops working when you change the table to a schemaful. And it somehow works if you change the vars to an array of maps instead of structs, correct?
Let me check what's going on.

mumoshu:
Hi @nickchomey! I figured out that it was due to `omitempty`. Please see the additional code comment I've added in [fed2803](https://github.com/surrealdb/surrealdb.go/pull/227/commits/fed28039b4e80280d4f5df682f359664c86b80fd).

It worked for `map` because... we can't specify `omitempty` when it's a map. You know, but `omitempty` is a well-known field tag used not only for CBOR but also for JSON and YAML marshaling, and perhaps others in the Go ecosystem. It skips zero values when encoding, and the same applies to CBOR encoding; that's all there's to it.

However, it is worth noting, as I've done in this example, that it can sometimes look confusing! I was even confused, and had to look into CBOR encoded bytes of the RPC request struct to see what happened 🤦

mumoshu:
@nickchomey Thank you for the confirmation! Appreciate your continued support 🙏 

I've just pushed some fixes to the test cleanup logic so that it removes not only records in a table but all the fields and indices in it altogether (DELETE vs REMOVE).

I'm merging this and tackling #202 next. Fortunately, I already have something that works locally.
I'd appreciate it if you could give it a try, and see if it actually improves the performance of your app.

PR #235, opened 2025-07-18
Fix: Refactor SignIn and SignUp methods to accept a map for authentication…

CristhianAC:
This pull request updates the authentication methods in the `DB` struct and refactors related examples and tests to use a more flexible `map[string]any` type instead of the `Auth` struct. This change enables dynamic handling of authentication data and supports custom fields for various use cases.

### Refactoring Authentication Methods:

* [`db.go`](diffhunk://#diff-ac94c91f8ea84762d55b6932ef1466a2165984591d3e646360c5bb31b2130900L94-R94): Updated the `SignUp` and `SignIn` methods to accept `authData` as a pointer to `map[string]any` instead of the `Auth` struct, allowing for more flexible authentication data handling. [[1]](diffhunk://#diff-ac94c91f8ea84762d55b6932ef1466a2165984591d3e646360c5bb31b2130900L94-R94) [[2]](diffhunk://#diff-ac94c91f8ea84762d55b6932ef1466a2165984591d3e646360c5bb31b2130900L108-R108)

### Updates to Tests:

* [`db_test.go`](diffhunk://#diff-842765a20a0a937a8eb5267b53fb11b7493cc6c2a6a8871e1da2b684702c5bc9L105-R107): Modified the `signIn` helper function to use `map[string]any` for authentication data in test cases.

### Updates to Examples:

* [`example/connection.go`](diffhunk://#diff-23c32877e6097ebe56121b88575497bc2ccf5e47e9ab8126a02ec677a08397bcL55-R57): Refactored the `initConnection` function to use `map[string]any` for authentication data, replacing the `Auth` struct.
* [`example/main.go`](diffhunk://#diff-1c171c5d51ba71728458fc771ff98395bcd3f59481e736c18e059372723acaabL25-R27): Updated the `authData` structure in the `main` function to use `map[string]any` for signing in, enabling custom field names.
* [`example/record_user_test.go`](diffhunk://#diff-fe12ddc4fa5c87654b82e2e26f8e22ed53ffbc298935b0768eca88b75f54a9a6L49-R54): Refactored `ExampleDB_recordAuthentication` to use `map[string]any` for both `SignUp` and `SignIn` calls, supporting dynamic authentication fields. [[1]](diffhunk://#diff-fe12ddc4fa5c87654b82e2e26f8e22ed53ffbc298935b0768eca88b75f54a9a6L49-R54) [[2]](diffhunk://#diff-fe12ddc4fa5c87654b82e2e26f8e22ed53ffbc298935b0768eca88b75f54a9a6L65-R70)… data

### DIFF
--- a/db.go
+++ b/db.go
@@ -109,9 +109,9 @@ func (db *DB) Info() (map[string]interface{}, error) {
 // Example with map:
 //
 //	db.SignUp(map[string]any{
-//	  "namespace": "app",
-//	  "database": "app",
-//	  "scope": "user",
+//	  "NS": "app",
+//	  "DB": "app",
+//	  "AC": "user",
 //	  "user": "yusuke",
 //	  "pass": "VerySecurePassword123!",
 //	})
--- a/db.go
+++ b/db.go
@@ -2,7 +2,6 @@ package surrealdb
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"log/slog"
 	"net/url"
@@ -95,33 +94,30 @@ func (db *DB) Info() (map[string]interface{}, error) {
 //
 // The authData parameter can be either:
 //   - An Auth struct
-//   - A map[string]any with keys like: "NS", "DB", "SC", "user", "pass"
+//   - A map[string]any with keys like: "namespace", "database", "scope", "user", "pass"
 //
 // Example with struct:
-//   db.SignUp(Auth{Namespace: "app", Database: "app", Scope: "user", Username: "foo", Password: "bar"})
+//
+//	db.SignUp(Auth{
+//	  Namespace: "app",
+//	  Database: "app",
+//	  Scope: "user",
+//	  Username: "yusuke",
+//	  Password: "VerySecurePassword123!",
+//	})
 //
 // Example with map:
-//   db.SignUp(map[string]any{"NS": "app", "DB": "app", "SC": "user", "user": "foo", "pass": "bar"})
+//
+//	db.SignUp(map[string]any{
+//	  "namespace": "app",
+//	  "database": "app",
+//	  "scope": "user",
+//	  "user": "yusuke",
+//	  "pass": "VerySecurePassword123!",
+//	})
 func (db *DB) SignUp(authData interface{}) (string, error) {
-	var data map[string]any
-
-	switch v := authData.(type) {
-	case *map[string]any:
-		data = *v
-	case map[string]any:
-		data = v
-	default:
-		b, err := json.Marshal(v)
-		if err != nil {
-			return "", fmt.Errorf("error serializing auth data: %w", err)
-		}
-		if err := json.Unmarshal(b, &data); err != nil {
-			return "", fmt.Errorf("error deserializing auth data to map: %w", err)
-		}
-	}
-
 	var token connection.RPCResponse[string]
-	if err := db.con.Send(&token, "signup", data); err != nil {
+	if err := db.con.Send(&token, "signup", authData); err != nil {
 		return "", err
 	}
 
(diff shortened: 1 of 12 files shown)


CristhianAC:
This pull request introduces improvements to the authentication methods in the `surrealdb` package, enhancing flexibility by allowing multiple data formats for `SignUp` and `SignIn`. Additionally, it updates examples to reflect these changes. 

### Authentication Method Enhancements:

* **Refactored `SignUp` method**: Updated to accept either an `Auth` struct or a `map[string]any` for authentication data. This increases flexibility and supports dynamic data formats. Added serialization and deserialization logic for non-map inputs.
* **Refactored `SignIn` method**: Similar to `SignUp`, now accepts both `Auth` structs and `map[string]any`. Includes serialization and deserialization logic for broader input compatibility.

### Example Updates:

* **Modified `ExampleDB_recordAuthentication`**: Updated example to use `map[string]any` instead of `Auth` struct, demonstrating the new flexibility in authentication methods.

### Dependency Updates:

* **Added `encoding/json` import**: Required for serialization and deserialization of authentication data.

mumoshu (commented):
Thanks! This is very interesting indeed!

To the best of my knowledge, 

- you shouldn't need to do `&map[string]any{` because they are pass-by-reference anyway- the pointer indirection is unnecessary.`case *map[string]any` would be eliminated, and the example should use `map[string]any` instead of `&map[string]any`
- The example map should use lower case chars like `"namespace"` rather than `"Namespace"`
- As the example you touched had been using the Auth struct, the mapping needs to be tweaked like `"user": "yusuke"` and `"pass": "VerySecurePassword123!"` instead of `"Username": "yusuke"` and so on. The struct field `Username` had been converted to a CBOR map field of `"user"`, for example.
- `db.con.Send` should accept anything that you can marshal using CBOR. It doesn't necessarily be a `map[string]any` so I believe we can just pass anything to it, without redundant json xmarshaling and unmarshaling

CristhianAC:
I already implemented this

> Thanks! This is very interesting indeed!
> 
> To the best of my knowledge,
> 
> * you shouldn't need to do `&map[string]any{` because they are pass-by-reference anyway- the pointer indirection is unnecessary.`case *map[string]any` would be eliminated, and the example should use `map[string]any` instead of `&map[string]any`
> * The example map should use lower case chars like `"namespace"` rather than `"Namespace"`
> * As the example you touched had been using the Auth struct, the mapping needs to be tweaked like `"user": "yusuke"` and `"pass": "VerySecurePassword123!"` instead of `"Username": "yusuke"` and so on. The struct field `Username` had been converted to a CBOR map field of `"user"`, for example.
> * `db.con.Send` should accept anything that you can marshal using CBOR. It doesn't necessarily be a `map[string]any` so I believe we can just pass anything to it, without redundant json xmarshaling and unmarshaling

db.go
```diff
@@ -104,8 +128,32 @@ func (db *DB) SignUp(authData *Auth) (string, error) {
...
+//	})
+//
+// Example with map:
+//
+//	db.SignIn(map[string]any{
+//	  "namespace": "app",
+//	  "database": "app",
+//	  "scope": "user",
+//	  "user": "yusuke",
+//	  "pass": "VerySecurePassword123!",
```
mumoshu, line 154:
```suggestion
//	  "NS": "app",
//	  "DB": "app",
//	  "user": "yusuke",
//	  "pass": "VerySecurePassword123!",
```

as per https://surrealdb.com/docs/surrealdb/integration/rpc#parameters-20

db.go
```diff
@@ -104,8 +128,32 @@ func (db *DB) SignUp(authData *Auth) (string, error) {
...
+// The authData parameter can be either:
+//   - An Auth struct
+//   - A map[string]any with keys like: "namespace", "database", "scope", "user", "pass"
+//
+// Example with struct:
+//
+//	db.SignIn(Auth{
+//	  Namespace: "app",
+//	  Database: "app",
+//	  Scope: "user",
```
mumoshu, line 142:
```suggestion
//	  Access: "user",
```

db.go
```diff
@@ -90,8 +90,32 @@ func (db *DB) Info() (map[string]interface{}, error) {
...
+//	  Username: "yusuke",
+//	  Password: "VerySecurePassword123!",
+//	})
+//
+// Example with map:
+//
+//	db.SignUp(map[string]any{
+//	  "namespace": "app",
+//	  "database": "app",
+//	  "scope": "user",
```
mumoshu, line 114:
```suggestion
//	  "NS": "app",
//	  "DB": "app",
//	  "AC": "user",
```

db.go
```diff
@@ -90,8 +90,32 @@ func (db *DB) Info() (map[string]interface{}, error) {
...
+// The authData parameter can be either:
+//   - An Auth struct
+//   - A map[string]any with keys like: "namespace", "database", "scope", "user", "pass"
+//
+// Example with struct:
+//
+//	db.SignUp(Auth{
+//	  Namespace: "app",
+//	  Database: "app",
+//	  Scope: "user",
```
mumoshu, line 104:
```suggestion
//	  Namespace: "app",
//	  Database: "app",
//	  Access: "user",
```

mumoshu (approved):
LGTM. Thank you very much for your contribution @CristhianAC!

PR #249, opened 2025-07-20
Fix: don't panic on invalid record ID

martinbockt:
### What does this change?
* Replaces the `panic` in `ParseRecordID` with a `(val, error)` return.
* Introduces a sentinel error `ErrBadRecordID` so callers can `errors.Is(..)` without relying on string matching.

### Why?
1. Invalid user input *shouldn’t* crash the entire application—only the request that supplied it.
2. A panic exposed an easy denial-of-service vector: any request with an invalid `id` could terminate the process.
3. Returning an error follows Go’s standard error-handling idioms and makes unit-testing simpler.

### How to review
* **Focus file:** `pkg/models/record_id.go`
* Call-site updates are mechanical (`_, err := ParseRecordID(...)`).

### Backwards compatibility
This change is source-breaking: callers now receive `(val, error)`. They’ll need to add an error check, but that’s a compile-time failure—far better than the previous runtime panic on invalid input.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -174,7 +174,9 @@ func (s *SurrealDBTestSuite) TestInsert() {
 }
 
 func (s *SurrealDBTestSuite) TestPatch() {
-	_, err := surrealdb.Create[testUser](s.db, *models.ParseRecordID("users:999"), map[string]interface{}{
+	recordID, err := models.ParseRecordID("users:999")
+	s.Require().NoError(err)
+	_, err = surrealdb.Create[testUser](s.db, *recordID, map[string]interface{}{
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
@martinbockt Thank you for your contribution! Good catch, we should do this!

My notes to everyone:

- Have you been using this for your application code? Please handle the new error!
- Used this in tests? Create a helper like `MustParseRecordID(string)` that wraps the enhanced `ParseRecordID`.

mumoshu:
@martinbockt FYI, this is now released as a part of v0.5.0 👋

Issue #144, opened 2024-08-13
Consider switching WebSocket library


JohnKinyanjui:
### Describe the bug

Am not sure but i think this error might by gorilla websocket please consider to switch to a better library, it unable to handle websocket correctly but if its from my end let me know

### Steps to reproduce

this is my simple middleware in golang for echo, i use it to authorize tokens on the rest api

```go
func AuthMiddleware(next echo.HandlerFunc) echo.HandlerFunc {

	return func(c echo.Context) error {
		token := c.Request().Header.Get("Authorization")
		tokenString := strings.Split(token, " ")
		if len(tokenString) != 2 {
			return c.JSON(http.StatusUnauthorized, map[string]string{"message": "missing or invalid token"})
		}

		db, err := database.Connection(tokenString[1])
		if err != nil {
			return c.JSON(http.StatusUnauthorized, map[string]string{"message": "invalid token"})
		}
		defer db.Close()

		c.Set("db", db)

		return next(c)
	}
}

```

### Expected behaviour

```bash
panic: repeated read on failed websocket connection

goroutine 315 [running]:
github.com/gorilla/websocket.(*Conn).NextReader(0xc000557b80)
        C:/Users/hp/go/pkg/mod/github.com/gorilla/websocket@v1.5.1/conn.go:1055 +0x28d
github.com/gorilla/websocket.(*Conn).ReadMessage(0xc00025ff70?)
        C:/Users/hp/go/pkg/mod/github.com/gorilla/websocket@v1.5.1/conn.go:1120 +0x13
github.com/surrealdb/surrealdb.go/internal/websocket.(*WebSocket).read(0xc0005b04e0?, {0x12f51e0, 0xc0003359e0})
        C:/Users/hp/go/pkg/mod/github.com/surrealdb/surrealdb.go@v0.2.1/internal/websocket/ws.go:140 +0x25
github.com/surrealdb/surrealdb.go/internal/websocket.(*WebSocket).initialize.func1()
        C:/Users/hp/go/pkg/mod/github.com/surrealdb/surrealdb.go@v0.2.1/internal/websocket/ws.go:167 +0x5e
created by github.com/surrealdb/surrealdb.go/internal/websocket.(*WebSocket).initialize in goroutine 351
        C:/Users/hp/go/pkg/mod/github.com/surrealdb/surrealdb.go@v0.2.1/internal/websocket/ws.go:160 +0x4f
```

### SurrealDB version

1.5.4 for windows on x86_64

[...]

Issue #110, opened 2023-10-29
Bug: Re-connection to the DB


iliesh:
### Describe the bug

When the application is running, and the DB is restarted, then the application panic with the following stack:
```
panic: repeated read on failed websocket connection

goroutine 50 [running]:
github.com/gorilla/websocket.(*Conn).NextReader(0xc0000e0580)
        /home/admin/go/pkg/mod/github.com/gorilla/websocket@v1.5.0/conn.go:1030 +0x4f4
github.com/gorilla/websocket.(*Conn).ReadMessage(0x129dfa0?)
        /home/admin/go/pkg/mod/github.com/gorilla/websocket@v1.5.0/conn.go:1093 +0x2f
github.com/surrealdb/surrealdb.go/internal/websocket.(*WebSocket).read(0xc00039c080, {0xca5720, 0xc0003f4480})
        /home/admin/go/pkg/mod/github.com/surrealdb/surrealdb.go@v0.2.1/internal/websocket/ws.go:140 +0x45
github.com/surrealdb/surrealdb.go/internal/websocket.(*WebSocket).initialize.func1()
        /home/admin/go/pkg/mod/github.com/surrealdb/surrealdb.go@v0.2.1/internal/websocket/ws.go:167 +0xac
created by github.com/surrealdb/surrealdb.go/internal/websocket.(*WebSocket).initialize in goroutine 1
        /home/admin/go/pkg/mod/github.com/surrealdb/surrealdb.go@v0.2.1/internal/websocket/ws.go:160 +0x8e
exit status 2
```

### Steps to reproduce

run an application and then stop surrealdb

### Expected behaviour

reconnect or return back an error to the application, and let developer decide what to do

### SurrealDB version

1.0.0+20230913.54aedcd for linux on x86_64

### Contact Details

iliusha.md@gmail.com

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ElecTwix:
Thanks for the issue.
As you said It needs to reconnect after the websocket connection is broken after other PRs merged I will look into this.

phughk:
We do want a shared configuration between drivers about how we handle network traffic (timeouts, reconnects, circuiting breaking strategies etc). Overall this would need to be standardised but for now we could add a helper.

Would recommend adhering to existing golang standards.

mumoshu:
Thank you @0xCA!

[...]

PR #269, opened 2025-07-27
Add support for gws auto-reconnection

mumoshu:
This is a follow-up to our new gws-based WebSocket connection (issue #144, PR #268), which essentially repeats #110 that we previously added to our standard `gorilla/websocket` mode, but to the new gws-based connection this time.

As I've mentioned in #268, this also enhances our CI matrix to cover gws and reconnecting-gws tests.

To be clear, our existing tests do not actually test scenarios that involves reconnecting, because we don't currently have a reliable and fast way to simulate disconnection.

Maybe I will try to write a fake WebSocket server that speaks limited SurrealDB RPC with failure injection supports(like disconnection, response delays, etc) and add a few test cases that leverages it, so that we can be sure that the reconnection is working.


0xCA:
testcontainers allow to easily run tests locally too, independent of GH actions

PR #271, opened 2025-07-27
Faster rand.NewRequestID

0xCA:
Basically, this. I ported my code from sdbc here, and improved it further slightly.

Benchmark results:
```
BenchmarkNewRequestID1-16         647862              1778 ns/op              32 B/op          2 allocs/op
BenchmarkNewRequestID2-16         558726              1933 ns/op              64 B/op          3 allocs/op
BenchmarkNewRequestID3-16         911889              1271 ns/op              32 B/op          2 allocs/op
BenchmarkNewRequestID4-16         828955              1406 ns/op              32 B/op          2 allocs/op
BenchmarkNewRequestID5-16         714678              1609 ns/op              32 B/op          2 allocs/op
```
Version 1 is the original, Version 3 is used in actual request methods. The rest is for comparison. See code comments for details.

> ### WARNING: I did bump Go version to 1.22 in go.mod. If it's not acceptable, I can re-write the code to compile on 1.20.

### DIFF
--- a/internal/rand/rand.go
+++ b/internal/rand/rand.go
@@ -1,27 +1,92 @@
 package rand
 
 import (
-	"math/rand"
+	cryptorand "crypto/rand"
+	"encoding/binary"
+	"math/rand/v2"
 	"sync"
-	"time"
 )
 
-const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+const (
+	bytesInUint64 = 8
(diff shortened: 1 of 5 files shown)


mumoshu (approved):
Good catch, and LGTM! Thank you for your contribution @0xCA!

I even realized this does not need to be a random ID now. A unique ID would be enough, like `unique client ID` concatenated with `monotonic ID`. But that's another story...

mumoshu:
@0xCA Could you rebase your changes? I recently moved pkg/connection/ws.go to `pkg/connection/gorillaws/ws.go`

0xCA:
@mumoshu sure! Rebase done

PR #333, opened 2025-08-29
Documentation overhaul

mumoshu:
Since the next minor release of this SDK (v0.10.0), I'd like to use pkg.go.dev as the most up-to-date, easy-to-browse, and comprehensive documentation for any released versions of this SDK.

Today, you can see that it lacks examples associated with each top-level symbol:

https://pkg.go.dev/github.com/surrealdb/surrealdb.go

However, you can see function docs and testable examples side by side since #329, #330, and #331.

Now the remaining part is that our current README lacks proper links to each top-level symbol and its Go doc comment.
In this pull request, I changed that by moving most of the detailed documentation to go doc comments, allowing you to do everything in a more structured way on pkg.go.dev.

Once the v0.10.0 gets published, you will be able to compare the enhancements by browsing v0.9.0 version of our pkg.go.dev page (https://pkg.go.dev/github.com/surrealdb/surrealdb.go@v0.9.0) vs one for v010.0 (https://pkg.go.dev/github.com/surrealdb/surrealdb.go@v0.10.0, not available just now because we cut v0.10.0 after this pull request).

> Note that you still need to read the README.md files and go through the doc comments in the code if you want to use unreleased versions. That's unchanged.
>
> I believe it's less common though.
>
> I believe most of the folks contacted me for a lack of documentation/confusing documentation
> referred to the one on surrealdb.com. It will be overhauled via https://github.com/surrealdb/docs.surrealdb.com/pull/1422 separately.
> But in the future, we could derive docs on surrealdb.com from the one published at pkg.go.dev if it's easier overall.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
     <img width=120 src="https://raw.githubusercontent.com/surrealdb/icons/main/golang.svg" />
 </p>
 
-<h3 align="center">The official SurrealDB SDK for Golang.</h3>
+<h3 align="center">The official SurrealDB SDK for Go</h3>
 
 <br>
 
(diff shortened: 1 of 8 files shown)


doc.go
```diff
@@ -0,0 +1,32 @@
+// The [surrealdb] package implements [SurrealDB RPC Protocol] in the Go way.
```
mumoshu, line 1: This will appear above the function index at https://pkg.go.dev/github.com/surrealdb/surrealdb.go#section-documentation so that the reader can go through (1) README and (2) package doc, to (3) each essential function doc in this order.

README.md
```diff
@@ -30,281 +30,19 @@
...
-```
-
-For Memory
-
-```go
-db, err := surrealdb.New("mem://")
-db, err := surrealdb.New("memory://")
-```
-
-## Data Models
```
mumoshu, line 231: Moved to the `models` package README, so that it appears in https://pkg.go.dev/github.com/surrealdb/surrealdb.go/pkg/models, making the documentation more structured and therefore easier to consume.

README.md
```diff
@@ -30,281 +30,19 @@
...
-| Geometry Multipoint              | `surrealdb.GeometryMultiPoint{GeometryPoint1, GeometryPoint2,... }`                                         |                                                                                        |
-| Geometry MultiLine               | `surrealdb.GeometryMultiLine{GeometryLine1, GeometryLine2,... }`                                            |                                                                                        |
-| Geometry MultiPolygon            | `surrealdb.GeometryMultiPolygon{GeometryPolygon1, GeometryPolygon2,... }`                                   |                                                                                        |
-| Geometry Collection              | `surrealdb.GeometryMultiPolygon{GeometryPolygon1, GeometryLine2, GeometryPoint3, GeometryMultiPoint4,... }` |                                                                                        |
-
-## Helper Types
-
-### surrealdb.O
```
mumoshu, line 268: Removed because it does not exist. Apparently we have `surrealdb.Obj` but I deprecated it in this pull request because it does't make much sense. You could use `map[string]any` instead of `surrealdb.Obj`- the former is 14 chars while the latter is 13 chars, but we won't need a type alias just to save one char.

README.md
```diff
@@ -30,281 +30,19 @@
...
-})
-```
-
-Where surrealdb.O is defined below. There is no special advantage in using this other than simplicity/legibility.
-
-```go
-type surrealdb.O map[any]any
-```
-
-### surrealdb.Result[T]
```
mumoshu, line 296: Removed because it was inappropriate. You should use `surrealdb.RPCResponse` to receive RPC responses.

README.md
```diff
@@ -30,281 +30,19 @@
...
-db, err := surrealdb.FromEndpointURLString(ctx, "http://localhost:8000")
-```
-
-or for a secure connection
-
-```go
-db, err := surrealdb.FromEndpointURLString(ctx, "https://localhost:8000")
-```
-
-### Using SurrealKV and Memory
```
mumoshu, line 207: Removed because they aren't implemented.

README.md
```diff
@@ -30,281 +30,19 @@
...
-You can run any of the example tests with:
-```sh
-go test -v ./ -run ExampleName
-```
-
-If you're viewing this documentation at `pkg.go.dev`, you can view the examples alongside corresponding SDK functions.
-
-## Executing SurrealQL
-
-### Using Query
```
mumoshu, line 87: Moved to the `surrealdb` package doc and to the `surrealdb.Query` doc.

surrealdb package: https://pkg.go.dev/github.com/surrealdb/surrealdb.go#section-documentation
Query: https://pkg.go.dev/github.com/surrealdb/surrealdb.go#Query

README.md
```diff
@@ -30,281 +30,19 @@
     <a href="https://www.youtube.com/channel/UCjf2teVEuYVvvVC-gFZNq6w"><img src="https://img.shields.io/badge/youtube-subscribe-fc1c1c.svg?style=flat-square"></a>
 </p>
 
-# surrealdb.go
-
-The official SurrealDB SDK for Go.
```
mumoshu, line 35: Looked like a repetition of the header above. Removed.

README.md
```diff
@@ -30,281 +30,19 @@
...
-
-## Documentation
```
mumoshu, line 37: Turned into `# Links`. Please see the new Links section markdown comments for more context.

types.go
```diff
@@ -81,8 +83,10 @@ type Auth struct {
 	Password  string `json:"pass,omitempty"`
 }
 
+// Deprecated: Use map[string]any instead
```
mumoshu, line 86: Deprecated as mentioned previously. Let's remove this when we cut v1.

PR #341, opened 2025-09-04
Remove NewGeometryPoint / Reorder Latitude and Longitude in GeometryPoint

nickchomey:
follow-on from #225, swap position in Struct to maintain consistency


mumoshu:
@nickchomey Great! Could you remove it along with fixing the tests?

Or I'll do that later, if you don't mind me pushing some commits to your fork!

mumoshu:
@nickchomey Thank you for your confirmation!
I removed the constructor and fixed the tests.
I've also repurposed the potential-breaking-change doc to become a v1 migration guide.
This should be good to go now.

Thank you very much for your contribution!

PR #343, opened 2025-09-04
Fix datetime marshal to allow extreme values

nickchomey:
previous mechanism was trimming values at the extremes. This doesn't and is simpler as it just extracts the seconds and nanoseconsd directly from the time.Time

### DIFF
--- a/pkg/models/datetime.go
+++ b/pkg/models/datetime.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	"github.com/fxamacker/cbor/v2"
-	"github.com/surrealdb/surrealdb.go/pkg/constants"
 )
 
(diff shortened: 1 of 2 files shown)


mumoshu:
Thank you, @nickchomey! Ah, looks great! I will take a closer look at this and CI failures more carefully after #344.

FWIW, I found a similar (in some sense) issue in `CustomDuration` too, although it was more about precision loss due to float<->int conversion with extreme values...
https://github.com/surrealdb/surrealdb.go/blob/af169df9d12dd63ffcba32324d566cc4adda9d02/pkg/models/duration.go#L72
which I discovered and fixed when integrating those custom types unmarshaler to `surrealcbor` in #344.

PR #347, opened 2025-09-15
Fix db tests.

ssttuu:
1. Updates the "create" test to use the `Create` method rather than `Select`.
2. Add a few more expectations.


mumoshu (approved):
Awesome! Thanks for fixing it, @ssttuu!

PR #363, opened 2025-10-07
surrealql: add String() method to ForStatement to return SurrealQL st…

### LINKED ISSUE #362: Bug: surrealql.For does not have String method
### Describe the bug

When trying to use `.For` inside a Query (e.g to use it on a transaction)

Example of error:
```
cannot use surrealql.For("X", X) (value of type *surrealql.ForStatement) as surrealql.Query value in argument to surrealql.Begin().Query: *surrealql.ForStatement does not implement surrealql.Query (missing method String)
``

### Reproduction

```go
 surrealql.Begin().
		Query(surrealql.For("var", "?", variable))
```

### Steps to reproduce

Initialize a Begin statement and then use a Query with a Fot

### Expected behaviour

`Query` method to accept ForStatement `ForStatement`

### Error

cannot use surrealql.For("offer", "?", offers) (value of type *surrealql.ForStatement) as surrealql.Query value in argument to surrealql.Begin().Query: *surrealql.ForStatement does not implement surrealql.Query (missing method String)

### SDK version

v1.0.0

### SurrealDB version

2.3.7

### Contact Details

_No response_

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Sergionx:
Thix fixes #362

### DIFF
--- a/contrib/surrealql/for.go
+++ b/contrib/surrealql/for.go
@@ -40,3 +40,9 @@ func (f *ForStatement) Build() (sql string, vars map[string]any) {
 
 	return builder.String(), c.vars
 }
+
+// String returns the SurrealQL string
+func (q *ForStatement) String() string {
+	sql, _ := q.Build()
+	return sql
+}


mumoshu (approved):
Awesome! Thank you very much for your contribution @Sergionx!

PR #375, opened 2026-01-29
feat: sanitize identifiers to prevent SQL injection in restore

RoseSecurity:
## Why

> [!NOTE]
> Feel free to adopt, disregard, or close this. My feelings won't be hurt. Just thought this might be useful!

Add a `sanitizeIdentifier` function to wrap SurrealDB identifiers in backticks and escape internal backticks. Apply this to all dynamic identifiers in restore.go to prevent SQL injection vulnerabilities when restoring data or defining schema elements.

## What

* Added a `sanitizeIdentifier` function to escape SurrealDB identifiers by wrapping them in backticks and escaping any internal backticks, preventing injection vulnerabilities.
* Updated all relevant query constructions in `restore.go` to use `sanitizeIdentifier` for table names, field names, namespaces, and database names, including in SET, INSERT, DEFINE NAMESPACE, DEFINE DATABASE, and DEFINE TABLE statements.
* Modified the `formatRecordID` function to sanitize table names when converting to the `table:id` format.

### DIFF
--- a/contrib/surrealrestore/restore.go
+++ b/contrib/surrealrestore/restore.go
@@ -227,7 +227,7 @@ func (r *Restorer) fullFromReader(ctx context.Context, currentNamespace, current
 						i := 0
 						for k, v := range record.Data {
 							paramName := fmt.Sprintf("param_%d", i)
-							setItems = append(setItems, fmt.Sprintf("%s = $%s", k, paramName))
+							setItems = append(setItems, fmt.Sprintf("%s = $%s", sanitizeIdentifier(k), paramName))
 							vars[paramName] = v
 							i++
 						}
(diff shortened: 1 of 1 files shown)


mumoshu (changes requested):
Hi @RoseSecurity! Thank you very much for your contribution!

I just double checked what your code intends and how SurrealDB queries work, and I got to think this won't work 🤔 

For example, you cannot say like:

```
DEFINE TABLE ```foo```
```

like the `sanitizeIdentifier` you've added assumes, to define a table whose name is surrounded by single backticks.

Instead, if I'm not misunderstanding it terribly, it rather needs to be:

```
DEFINE TABLE `\`foo\``
```

I'd appreciate your confirmation when you have time!

RoseSecurity:
Hi @mumoshu,

Thanks for the call out, my understanding was off, so I went back and read the SurrealDB parser source around the `lex_common_escape_sequence` in `surrealdb/core/src/syn/lexer/strings/mod.rs` which shows:

```rust
b'`' => buffer.push(b'`'),  // \` escapes a backtick
```

I've updated the implementation to escape backslashes first, then backticks:

```go
escaped := strings.ReplaceAll(id, "\\", "\\\\")
escaped = strings.ReplaceAll(escaped, "`", "\\`")
```

Thanks for catching this! If this isn't something you're interested in adding, feel free to disregard or close. I just wanted to contribute to your awesome tooling. If you ever want to collaborate on a Terraform provider for it, just say the word

mumoshu (approved):
@RoseSecurity LGTM! Thank you very much for your contribution!

JFYI, this is part of `surrealrestore`, one of the various demo projects this SDK provides. Any application code should better use `models.Table(tableName)` as a query variable instead! (Apparently, this was my oversight when I initially created surrealrestore, which you kindly fixed now!

I also realized that there was no direct example of using `models.Table` to escape table names, which has been added via #379. Please refer to the example, and all others, when you use this SDK to write your own apps!

Thank you again for your contribution!

PR #377, opened 2026-01-31
Add SignUpWithRefresh and SignInWithRefresh for WITH REFRESH Record Access Method in SurrealDB v3

mumoshu:
## What is the motivation?

SurrealDB v3 introduces the `WITH REFRESH` option for `TYPE RECORD` access methods, enabling refresh token support. This allows applications to obtain both an access token and a refresh token during sign-up/sign-in, and later use the refresh token to obtain new tokens without requiring the user to re-enter credentials.

The existing `SignUp` and `SignIn` methods only return a single access token string and cannot handle the two-token response format required for refresh token workflows.

## What does this change do?

This PR adds support for Record Access Methods with refresh tokens in SurrealDB v3:

**New API Methods:**
- `SignUpWithRefresh(ctx, authData)` - Signs up and returns both access and refresh tokens
- `SignInWithRefresh(ctx, authData)` - Signs in and returns both access and refresh tokens
- `Tokens` type alias exposing `Access` and `Refresh` fields

**Token Refresh Flow:**
```go
// Initial signin with credentials
tokens, err := db.SignInWithRefresh(ctx, map[string]any{
  "NS": "app", "DB": "app", "AC": "user_access",
  "user": "yusuke", "pass": "VerySecurePassword123!",
})

// Later: refresh tokens without credentials
newTokens, err := db.SignInWithRefresh(ctx, map[string]any{
  "NS": "app", "DB": "app", "AC": "user_access",
  "refresh": tokens.Refresh,
})

### CODE REVIEW

File: contrib/testenv/record_access_method_refresh_test.go [hunk 1]
```diff
@@ -0,0 +1,455 @@
...
+	err = connection.Send(conn, ctx, &queryResponse, "query", `RETURN "authenticated"`, nil)
+	require.NoError(t, err)
+
+	// 8. Verify refresh token can be used for SignIn (without username/password)
+	// The refresh token is passed via the "refresh" parameter (not "key")
+	refreshAuthData := map[string]any{
+		"NS":      "test_ns",
+		"DB":      "test_db",
+		"AC":      "user_access",
+		"refresh": refresh, // Use refresh token - no username/password needed
```
mumoshu, line 251: Ref https://github.com/surrealdb/surrealdb/blob/ef9d5608aa7b711993d3effe0f892d52e6529ea5/surrealdb/core/src/iam/signin.rs#L282-L294

PR #380, opened 2026-02-16
Error handling

kearfy:
## Structured error handling for SurrealDB v3

Adds full support for the structured error format introduced in SurrealDB v3.

### What changed

**V3 detail format support** — SurrealDB v3 switched from serde externally-tagged enums (`{"Auth": "TokenExpired"}`) to an internally-tagged format (`{"kind": "Auth", "details": {"kind": "TokenExpired"}}`). New helpers (`detailKind`, `detailInner`, `getDetailString`) handle the v3 format while retaining full backward compatibility with the old format.

**Recursive cause chain** — `ServerError` now carries a `cause *ServerError` field, matching Rust's `cause: Option<Box<Error>>`. The chain is parsed recursively from both RPC errors and query result errors. Since Go has no automatic error-chain display (unlike JS's `Error.cause` or Python's `__cause__`), `Error()` now joins the full cause chain with `": "` so the complete context is visible in logs. A new `Message()` method returns just the current error's message without the chain.

**Cause chain traversal** — `ServerCause()`, `Unwrap()`, `HasKind()`, and `FindCause()` methods on `*ServerError`, plus top-level `HasKind(err, kind)` and `FindCause(err, kind)` functions that work on any `error`.

**Kind-aware `errors.Is`** — `Is()` now supports optional kind matching: `errors.Is(err, &ServerError{})` matches any server error (catch-all), while `errors.Is(err, &ServerError{kind: "NotFound"})` only matches that specific kind.

**New detail accessors** — `SessionID()` and `TargetName()` for parity with the Python SDK. Updated doc comments on `NamespaceName()` and `DatabaseName()` to reflect they work for both `NotFound` and `AlreadyExists` kinds.

**Double-wrapped details unwrapping** — The server's query result path may double-wrap details (e.g. `{"kind": "AlreadyExists", "details": {"kind": "Record", ...}}`). `parseQueryError` now detects and unwraps this so detail accessors like `RecordID()` work correctly.

**Integration tests** — New `contrib/testenv/structured_errors_integration_test.go` validates structured errors against a live SurrealDB v3 instance: invalid credentials, parse errors, `THROW`, duplicate records, schema violations, and mixed multi-statement queries.

**CI** — Updated test matrix from `v3.0.0-beta.2` to `v3.0.0` GA.

[...]

PR #382, opened 2026-03-02
Add support for SurrealDB v3 structured error handling

mumoshu:
Introduce `ServerError` for SurrealDB v3 structured errors while preserving backward compatibility with the existing RPCError and QueryError types.

Keep using `RPCError` in SurrealDB v2 (although we recommend upgrading to v3, and' RPCError' has been deprecated to signal that).

Use `ServerError` on SurrealDB v3.

Errors embedded inside `QueryResult` remain as `QueryError`.

```go
_, err := surrealdb.Select[map[string]any](context.Background(), db,
	models.Table("nonexistent_table_for_test"),
)

// Use it on SurrealDB v3 only
var se surrealdb.ServerError
if errors.As(err, &se) {
	// Get access to new error fields like Kind, Details, and Cause
}

// You can also use this idiom to obtain the cause
var cause surrealdb.ServerError
if errors.As(errors.Unwrap(se), &cause) {
	// Get access Kind, Details, and Cause of Cause
}
```

Based on #380

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,12 +16,12 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # v3.0.0-beta.2 with Go 1.25.0
-          - surrealdb-version: 'v3.0.0-beta.2'
+          # v3.0.1 with Go 1.25.0
+          - surrealdb-version: 'v3.0.1'
             go-version: '1.25.0'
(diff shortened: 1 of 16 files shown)


kearfy (approved):

mumoshu:
Thanks for your contribution, @kearfy!

PR #387, opened 2026-03-19
Cleanup and update dependencies

WolfieLeader:
## Summary

Housekeeping: remove a stale config backup and bump two outdated dependencies.

## Commits

1. **Remove `.golangci.bck.yml`** — leftover from golangci-lint v1→v2 migration, references deprecated linters (`exportloopref`, `gomnd`), unreferenced by any workflow or Makefile
2. **Bump `fxamacker/cbor/v2` v2.7.0→v2.9.0** — adds opt-in TextMarshaler/JSON transcoding, `omitzero` struct tag. No API changes; all `EncMode`/`DecMode`/`TagSet`/`RawMessage` usage unchanged
3. **Bump `stretchr/testify` v1.8.4→v1.10.0** — test-only dep, adds new assertion helpers (`NotElementsMatch`, `NotErrorAs`, in-order mock validation). No breaking changes

## Compatibility

- No source code changes required for either dependency bump
- Go minimum unchanged at 1.23 (`cbor` v2.8+ needs 1.20, `testify` needs 1.17)
- Tested with both `surrealcbor` and `fxamackercbor` codec paths

## Testing

- `make lint` — 0 issues
- `make build` — compiles cleanly
- `make test` — passes with all codec/connection combinations

PR #388, opened 2026-03-19
Add test coverage for untested DB methods

WolfieLeader:
## Summary

Add suite test coverage for 5 public DB methods that previously had only example tests: `Version`, `Let`/`Unset`, `Upsert`, `Invalidate`.

## Motivation

These methods are part of the public API but lacked proper test assertions in the test suite. Example tests verify they don't panic but don't validate return values or behavior. Main package coverage increases from 57.7% to 58.6%.

## Tests Added

- **TestVersion** — verifies `Version()` returns non-nil `VersionData`
- **TestLetUnset** — verifies `Let()` sets variables readable via `RETURN $var` query, `Unset()` removes them so query returns nil
- **TestUpsert** — verifies create-if-not-exists and update-if-exists behavior with `RecordID` targets
- **TestInvalidate** — verifies session invalidation succeeds, re-signs in to avoid affecting other tests

## Not Included

- **`Info()`** — panics with nil pointer dereference when signed in as root (`db.go:152` dereferences `*info.Result` without a nil check, but the RPC returns null for root). This appears to be a bug — will report separately.
- **`SignUp`/`SignUpWithRefresh`/`Authenticate`** — require complex setup (RECORD access methods, JWT keys). Better addressed in a separate PR.

## Testing

- `make lint` — 0 issues
- `make build` — compiles cleanly
- `make test` — passes (same baseline as main)
- All new tests pass individually and in the full suite

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -527,6 +527,74 @@ func (s *SurrealDBTestSuite) TestQueryRaw() {
 	s.Require().NoError(err)
 }
 
+func (s *SurrealDBTestSuite) TestVersion() {
+	s.Run("returns version data", func() {
+		ver, err := s.db.Version(context.Background())
+		s.Require().NoError(err)
+		s.Require().NotNil(ver)
+	})
+}
+
+func (s *SurrealDBTestSuite) TestLetUnset() {
+	s.Run("let sets variable readable via query", func() {
+		err := s.db.Let(context.Background(), "testvar", "hello")
+		s.Require().NoError(err)
+
+		result, err := surrealdb.Query[string](context.Background(), s.db, "RETURN $testvar", nil)
+		s.Require().NoError(err)
+		s.Require().NotNil(result)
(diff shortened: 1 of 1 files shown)


itsezc:
Hi @WolfieLeader, thanks for this PR, do you mind addressing the failing tests on the workflows. 

Thanks!

WolfieLeader:
Yeah sure, Ot would take some time though

WolfieLeader:
Damn it did took too much time

WolfieLeader:
@itsezc I will fix now

WolfieLeader:
Done for both of my PRs also look closely, I updated a major
#388 and #387

WolfieLeader:
There is also a test related PR which is #410 with a one liner fix

WolfieLeader:
Noice!

mumoshu (approved):
LGTM.
Thank you very much for your patience and contribution once again @WolfieLeader!

PR #394, opened 2026-04-09
fix: prevent data race between Kill/CloseLiveNotifications and handleResponse

### LINKED ISSUE #393: Bug: Data race in Kill/CloseLiveNotifications
### Describe the bug

CloseLiveNotifications closes the notification channel while handleResponse (spawned by readLoop) may still be sending on it. This causes a data race detected by Go's -race detector and can panic with send on closed channel.

### Reproduction

Run any test that creates and cancels a live query with go test -race. The race is intermittent but reproduces reliably under parallel test execution.

### Steps to reproduce

go.mod:

```go.mod
module race.detector

require (
      github.com/docker/docker v28.5.2+incompatible
      github.com/surrealdb/surrealdb.go v1.4.0
      github.com/testcontainers/testcontainers-go v0.41.0
  )
```

test.go:

```go
package main

import (
	"context"
	"fmt"
	"os"
	"testing"
	"time"

	"github.com/docker/docker/api/types/container"
	"github.com/surrealdb/surrealdb.go"
	"github.com/surrealdb/surrealdb.go/pkg/models"
	"github.com/testcontainers/testcontainers-go"
	"github.com/testcontainers/testcontainers-go/wait"
)

var endpoint string

func TestMain(m *testing.M) {
	ctx := context.Background()

	surreal, err := testcontainers.GenericContainer(ctx,
		testcontainers.GenericContainerRequest{
			ContainerRequest: testcontainers.ContainerRequest{
				Image: "surrealdb/surrealdb:v3.0.5",
				Env: map[string]string{
					"SURREAL_PATH": "memory",
					"SURREAL_USER": "root",
					"SURREAL_PASS": "root",
				},
				Cmd:          []string{"start", "--allow-funcs", "--log", "info"},
				ExposedPorts: []string{"8000/tcp"},
				WaitingFor:   wait.ForLog("Started web server on "),
				HostConfigModifier: func(conf *container.HostConfig) {
					conf.AutoRemove = true
				},
			},
			Started: true,
		},
	)
	if err != nil {
		fmt.Fprintf(os.Stderr, "failed to start container: %v\n", err)
		os.Exit(1)
	}

	ep, err := surreal.Endpoint(ctx, "")
	if err != nil {
		_ = surreal.Terminate(ctx)
		fmt.Fprintf(os.Stderr, "failed to get endpoint: %v\n", err)
		os.Exit(1)
	}
	endpoint = ep

	code := m.Run()

	_ = surreal.Terminate(ctx)
	os.Exit(code)
}

func TestLiveQueryKillRace(t *testing.T) {
	for i := range 50 {
		t.Run(fmt.Sprintf("iteration_%d", i), func(t *testing.T) {
			t.Parallel()

			ctx := context.Background()

			db, err := surrealdb.FromEndpointURLString(ctx, "ws://"+endpoint)
			if err != nil {
				t.Fatal(err)
			}
			defer func() { _ = db.Close(ctx) }()

			if _, err := db.SignIn(ctx, surrealdb.Auth{
				Username: "root",
				Password: "root",
			}); err != nil {
				t.Fatal(err)
			}

			ns := fmt.Sprintf("ns_%d", i)
			dbName := fmt.Sprintf("db_%d", i)
			if err := db.Use(ctx, ns, dbName); err != nil {
				t.Fatal(err)
			}

			if _, err := surrealdb.Query[any](ctx, db,
				"DEFINE TABLE item SCHEMALESS", nil,
			); err != nil {
				t.Fatal(err)
			}

			liveID, err := surrealdb.Live(ctx, db, models.Table("item"), false)
			if err != nil {
				t.Fatal(err)
			}

			notifications, err := db.LiveNotifications(liveID.String())
			if err != nil {
				t.Fatal(err)
			}

			for j := range 10 {
				if _, err := surrealdb.Query[any](ctx, db,
					fmt.Sprintf(`CREATE item SET value = %d`, j), nil,
				); err != nil {
					t.Fatal(err)
				}
			}

			select {
			case <-notifications:
			case <-time.After(5 * time.Second):
				t.Fatal("timeout waiting for notification")
			}

			if err := surrealdb.Kill(ctx, db, liveID.String()); err != nil {
				t.Logf("kill error (may be expected): %v", err)
			}
		})
	}
}
```

### Expected behaviour

Test runs without issue. No race condition appears.

### Error

WARNING: DATA RACE
Write at 0x00c000394e10 by goroutine 1227:
  runtime.closechan()
      runtime/chan.go:414 +0x0
  github.com/surrealdb/surrealdb.go/pkg/connection.(*Toolkit).CloseLiveNotifications()
      pkg/connection/connection.go:173 +0x12c

Previous read at 0x00c000394e10 by goroutine 1671:
  runtime.chansend()
      runtime/chan.go:176 +0x0
  github.com/surrealdb/surrealdb.go/pkg/connection/gorillaws.(*Connection).handleResponse()
      pkg/connection/gorillaws/connection.go:549 +0xc3d

### SDK version

v1.4.0

### SurrealDB version

v3.0.5

### Contact Details

_No response_

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Bahtya:
## Problem

Fixes #393

There is a data race between `CloseLiveNotifications`/`Kill` and `handleResponse` in WebSocket connections. When a live query is killed while notifications are still being received:

1. `Kill()` calls `CloseLiveNotifications(id)` which acquires the write lock, closes the notification channel, and removes it from the map
2. Concurrently, `handleResponse` (running in a separate goroutine) calls `GetNotificationChannel(id)` (read lock), gets the channel, releases the lock, then sends on the channel
3. Between releasing the read lock and sending, `CloseLiveNotifications` can close the channel, causing a **"send on closed channel" panic**

The race window exists because the read lock is released before the send operation.

## Fix

This PR introduces a `SendNotification` method on `Toolkit` that holds the read lock during the channel send, preventing `CloseLiveNotifications` from acquiring the write lock and closing the channel while a notification is being sent:

```go
func (bc *Toolkit) SendNotification(id string, notification Notification) bool {
    bc.NotificationChannelsLock.RLock()
    ch, ok := bc.NotificationChannels[id]
    if !ok {
        bc.NotificationChannelsLock.RUnlock()
        return false
    }

    select {
    case ch <- notification:
        bc.NotificationChannelsLock.RUnlock()
        return true
    default:
        bc.NotificationChannelsLock.RUnlock()
        return false
    }
}
```

Key design decisions:
- **Buffered channels (16)**: Notification channels are now created with a buffer of 16 to prevent the send from blocking while holding the read lock. This allows `SendNotification` to use a non-blocking send.
- **Drop-on-full**: If the channel buffer is full, the notification is dropped rather than blocking. This is safer than deadlocking and matches the real-time nature of live query notifications.
- **Both WS implementations updated**: Both `gorillaws` and `gws` `handleResponse` methods now use `SendNotification` instead of directly accessing the channel.

## Testing

- `go build ./...` passes
- `go vet ./pkg/connection/...` passes
- All unit tests pass (integration tests require a running SurrealDB instance)

---

*This PR was created by an AI contributor. The fix has been verified with build and vet checks.*

### DIFF
--- a/pkg/connection/connection.go
+++ b/pkg/connection/connection.go
@@ -110,7 +110,7 @@ func (bc *Toolkit) CreateNotificationChannel(liveQueryID string) (chan Notificat
 		return nil, fmt.Errorf("%w: %v", constants.ErrIDInUse, liveQueryID)
 	}
 
-	ch := make(chan Notification)
+	ch := make(chan Notification, 16)
 	bc.NotificationChannels[liveQueryID] = ch
 
 	return ch, nil
@@ -173,4 +173,29 @@ func (bc *Toolkit) CloseLiveNotifications(liveQueryID string) error {
 	close(ch)
 	delete(bc.NotificationChannels, liveQueryID)
 	return nil
+}
+
+// SendNotification safely sends a notification to the channel identified by id.
+// It holds the read lock during the send to prevent CloseLiveNotifications from
+// closing the channel concurrently, which would cause a "send on closed channel" panic.
+// Returns true if the notification was sent, false if the channel was not found
+// or was full (dropped).
+func (bc *Toolkit) SendNotification(id string, notification Notification) bool {
+	bc.NotificationChannelsLock.RLock()
+	ch, ok := bc.NotificationChannels[id]
+	if !ok {
+		bc.NotificationChannelsLock.RUnlock()
+		return false
+	}
+
+	select {
+	case ch <- notification:
+		bc.NotificationChannelsLock.RUnlock()
+		return true
+	default:
+		// Channel is full, drop the notification to avoid blocking
+		// while holding the read lock.
+		bc.NotificationChannelsLock.RUnlock()
+		return false
+	}
 }
--- a/pkg/connection/gorillaws/connection.go
+++ b/pkg/connection/gorillaws/connection.go
@@ -537,15 +537,11 @@ func (c *Connection) handleResponse(res []byte) {
 			return
 		}
 
-		LiveNotificationChan, ok := c.GetNotificationChannel(channelID.String())
-		if !ok {
+		if !c.SendNotification(channelID.String(), notification) {
 			c.logger.Error(
 				fmt.Sprintf("unavailable ResponseChannel %+v", channelID.String()),
 				"result", fmt.Sprint(rpcRes.Result),
 			)
-			return
 		}
-
-		LiveNotificationChan <- notification
 	}
 }
--- a/pkg/connection/gws/connection.go
+++ b/pkg/connection/gws/connection.go
@@ -146,12 +146,11 @@ func (c *Connection) handleResponse(data []byte) {
 			return
 		}
 
-		notificationChan, ok := c.GetNotificationChannel(notification.ID.String())
-		if !ok {
-			return
+		if !c.SendNotification(notification.ID.String(), notification) {
+			c.logError(
+				fmt.Sprintf("unavailable NotificationChannel %+v", notification.ID.String()),
+			)
 		}
-
-		notificationChan <- notification
 	}
 }
 


Bahtya:
Friendly ping 👋 This PR fixes a data race between Kill/CloseLiveNotifications and the notification goroutine. The fix uses proper synchronization. Would appreciate a review.

itsezc (approved):

PR #402, opened 2026-04-27
Handle killed notification

rushmorem:
This notification is not currently handled in this SDK which is causing panics in tests.


rushmorem:
@claude review

ssttuu (approved):

PR #410, opened 2026-06-29
Fix timezone-dependent time decoding in surrealcbor

WolfieLeader:
## Summary

Fix timezone-dependent `time.Time` decoding in the `surrealcbor` codec. Decoded
datetimes now have a canonical UTC location regardless of the offset in the
encoded value.

## Problem

The codec has two datetime decode paths that disagreed on location:

- **Tag 0 (RFC3339 string)** — `decode_tag.go` parsed with `time.Parse(time.RFC3339, str)`,
  which preserves the offset zone from the string (e.g. `+03:00`).
- **CustomDateTime (epoch `[seconds, nanoseconds]`)** — already normalized with `.UTC()`.

Because of this, a marshal/unmarshal round-trip produced a `time.Time` whose
`*time.Location` depended on the **producer's** timezone. `reflect.DeepEqual`
(used by `testify`'s `Equal`) compares the location, so round-trip equality
broke on any non-UTC machine. CI runs in UTC and never saw it; it reproduces
locally in a non-UTC zone:

```
--- FAIL: TestDecode_array_withAllSupportedTypes
    expected loc: (*time.Location)(<nil>)        // UTC
    actual   loc: (*time.Location)("Local", +03:00)
```

## Fix

Normalize the RFC3339 path to UTC, matching the epoch path — one line in
`decodeDateTimeStringTag`:

```go
t = t.UTC()
```

## Testing

- Reproduced the failure on a non-UTC (`TZ=IDT`) machine; confirmed fixed after the change.
- `go test ./...` green on both local (`TZ=IDT`) and `TZ=UTC`.
- `go vet ./...` and `golangci-lint run` clean.
- Affected tests: `TestDecode_array_withAllSupportedTypes`,
  `TestDecode_map_withAllSupportedTypes`,
  `TestDecode_map_structwithAllSupportedTypes_eachWithPointerAndNonPointerVariant`.


WolfieLeader:
Hey @mumoshu and @itsezc I suggest you to look into https://pullfrog.com/ by the creators of Zod and Arktype for PR code reviews, they also use it in their open source projects

mumoshu (approved):
Good catch! Thank you for fixing it @WolfieLeader!

WolfieLeader:
@mumoshu you have discord/linkedin or anything there is one small problem

PR #412, opened 2026-07-08
Fix Range/RecordRangeID.String() bugs

welpie21:
## Root cause

`pkg/models/range.go` had two provable bugs affecting the string rendering of `Range` and `RecordRangeID` values:

1. **Unimplemented stub.** `convertToString(v any) string` was a stub (`// todo: implement`) that always returned `""`. As a result, `Range.String()` and `RecordRangeID.String()` only ever rendered the bare join operator (e.g. `..`, `..=`, `>..`, `>..=`) — the actual bound values never appeared in the output.

2. **Begin/End copy-paste bug.** In both `Range.String()` (line ~94) and `RecordRangeID.String()` (line ~145), the end-bound string was computed by calling `convertToString(r.Begin)` / `convertToString(rr.Begin)` instead of passing the `End` field. So even once the stub is fixed, the rendered end bound would echo the begin value instead of the real end value whenever `Begin != End`.

## Fix

- Implemented `convertToString` to unwrap a `*BoundIncluded[T]`/`*BoundExcluded[T]` via reflection and stringify its `Value` field. String values are escaped the same way `RecordID.String()` escapes record IDs (reusing `needsEscaping`/`escapeString` from `record_id_string.go`), wrapping ambiguous values in angle brackets (e.g. `a-b` -> `⟨a-b⟩`). Other value types fall back to their default string form (`%v`).
- Fixed the `Begin`/`End` swap so the end bound is derived from the `End` field in both `String()` implementations.

The inclusive/exclusive join syntax (`..`, `..=`, `>..`, `>..=`) produced by `GetJoinString()` was already correct and was cross-checked against the golden reference `surrealdb.js` (`packages/sqon/src/internal/range.ts`'s `getRangeJoin` and `packages/sqon/src/utils/escape.ts`'s `escapeRangeBound`), which implement identical bound semantics.

## Tests

Added `pkg/models/range_test.go` covering:
- `Range.String()` and `RecordRangeID.String()` for all four inclusive/exclusive bound combinations (`..`, `..=`, `>..`, `>..=`)
- String bound escaping matching `RecordID.String()` conventions
- A regression case asserting the end bound reflects `End` and is not an echo of `Begin`
- `convertToString` directly, confirming it no longer returns `""`

I verified these tests fail against the pre-fix code (by stashing the `range.go` fix and re-running): `TestConvertToString` fails with empty-string results, confirming the stub was previously live and the new tests exercise the actual bug.

## Verification commands run

- `go build ./...` — passes
- `golangci-lint run ./...` — `0 issues.`
- `go test -v -run 'TestRange|TestRecordRangeID|TestConvertToString' ./pkg/models/...` — all pass
- `go test ./pkg/models/...` — `ok`

All tests are pure unit tests with no live SurrealDB server dependency.

## Scope / follow-up

This PR is narrowly scoped to the string-rendering correctness bugs above. It intentionally does **not** wire `RecordRangeID` into the `TableOrRecord` generic constraint so that ranges work end-to-end through `Select`/`Update`/`Delete`/`Upsert` — that CRUD-integration work is a separate, larger feature and is deferred to a follow-up PR.

No existing GitHub issue tracks this bug, so per this repo's `CONTRIBUTING.md` convention the branch name omits an issue-number prefix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/models/range.go
+++ b/pkg/models/range.go
@@ -91,7 +91,7 @@ func (r *Range[T, TBeg, TEnd]) String() string {
 		beginStr = convertToString(r.Begin)
 	}
 	if r.End != nil {
-		endStr = convertToString(r.Begin)
+		endStr = convertToString(r.End)
 	}
 
 	return fmt.Sprintf("%s%s%s", beginStr, joinStr, endStr)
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
LGTM! Awesome work!

